PR #5, opened 2019-05-08
Write CURIEs in `AbbreviatedIRI` XML elements instead of `IRI`

althonos:
Ok, so this one addresses a real issue: the current writer exports both IRIs and CURIEs into an `<IRI>` element whereas the specs expect an `<AbbreviatedIRI>` tag in the case of CURIEs. This causes Protégé to fail expanding CURIEs because it assumes the `<IRI>` element contains an expanded IRI already.

### DIFF
--- a/src/io/writer.rs
+++ b/src/io/writer.rs
@@ -71,7 +71,7 @@ fn attribute_maybe(elem: &mut BytesStart, key: &str, val: &Option<String>) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


phillord:
Yeah, I was making that up as I was going along. I didn't manage to find a clear definition of the distinction between IRI and abbreviatedIRI, but now I look again, there it is right at the start of the spec.

PR #7, opened 2019-05-14
Implement iterator traits for `Ontology` and `&Ontology` in `::model`

althonos:
The current model had no way of collecting all `AnnotatedAxiom`s in an `Ontology`. This PR adds public iterators (like in the `std::collections` module) that can be used to iterate on a borrowed or owned ontology.

This is also easier to use than current `impl Iterator` methods, which have the drawback of returning values that cannot be used in other structs, only consumed.

I also added `FromIterator<Into<AnnotatedAxiom>>` for `Ontology`.

### DIFF
--- a/benches/horned.rs
+++ b/benches/horned.rs
@@ -8,10 +8,11 @@ use bencher::Bencher;
 
 fn a_thousand_classes(bench: &mut Bencher) {
     bench.iter(|| {
+        let mut b = Build::default();
         let mut o = Ontology::new();
         for m in 1..1000 {
-            let i = o.iri(format!("http://example.com/b{}", m));
-            let _c = o.class(i);
+            let i = b.iri(format!("http://example.com/b{}", m));
+            let _c = o.declare(b.class(i));
         }
     })
 }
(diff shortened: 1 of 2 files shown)


phillord:
Why have you re-organised the files at the same time as changing things? Is the only functional change here, adding iterators?

althonos:
Yes, I just did that so the iterator structure (which is public) is exposed in `horned_owl::model::ontology::Iter`, like `std::collections::btree_set::Iter` is the iterator of `std::collections::BTreeSet` in the standard library.

phillord:
I think that `btree_set` does this because there are quite a few data structures all with an `Iter` struct. `BTreeMap`, by comparison does not. It would have been possible to add an struct called `Axioms` and implement `Iterator` over that. xml-rs does this (called `Events`). I don't really like whitespace changes where they can be avoided and in this case adding them into a single commit really confuses any diffs that we make.

althonos:
There are two iterators in here: `Iter` which borrows the annotated axioms, and `IntoIter` which takes ownership over them. And [`std::collections::btree_map::Iter` does exist](https://doc.rust-lang.org/std/collections/btree_map/index.html), as well as several other. Adding an `Axioms` structure could be confusing because the iteration could be owning or borrowing; in comparison, `std::collections::btree_map` does not have owning and borrowing `keys`, so there is no ambiguity. 

No namespace here is changed, since `Ontology` and `OntologyID` are both reexported in the `model` namespace, so it's not a breaking change at all!

phillord:
It's not the public API that is the problem, but you've combined a functional extension with a large white space change. It took me much longer to work out what the patch did because of it, and it will cause issues for any version control operation involving a diff.

So, happy to accept the iterator addition, but not the whitespace change.

althonos:
Rebased with everything in the same file.

phillord:
Can you kill the whitespace changes early in the file? I realise this is probalby just rustfmt, but because I didn't use rustfmt from the start, I think, we are limited to updating this as we go.

phillord:
Nearly there --- it needs some rebase/squashing. There need to be two or three commits here. One with iterator and one with the fixes to bench. The import fixes might go with iterator or not. Probably with I guess, but it doesn't make that much difference.

PR #8, opened 2019-05-18
Make sure to escape and unescape literal text in XML I/O code

althonos:
Current implementation does not decode and encode literal values, which is a pain for users since they are expected to do all the XML escaping themselves. This PR changes the implementation of `from_start! { Literal .. }` to use `read_text` so that the reader unescapes the literal value, and the writer was patched to expect an unescaped string in lieu of an escaped one.

I'm also not sure why `literal` is an `Option` in `Literal`, since the `None` type is here semantically equivalent to an empty string when serializing and deserializing.

~~*Also sorry for the large diff but I didn't realise I had `rustfmt` turned on in my IDE before pushing the commit*~~ Rebased and made a minimal commit without format changes.

### DIFF
--- a/src/io/reader.rs
+++ b/src/io/reader.rs
@@ -322,28 +322,17 @@ from_start! {
         let datatype_iri = read_a_iri_attr(r, e, b"datatypeIRI")?;
         let lang = attrib_value(r, e, b"xml:lang")?;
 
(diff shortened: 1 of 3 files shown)


phillord:
Didn't know `read_text` existed. Much simpler, thanks!

PR #10, opened 2019-05-18
Fix serialization/deserialization order of subclass axioms

### LINKED ISSUE #9: Subclass axioms are serialized/deserialized in the wrong order
The OWL XML2 specification states that in `SubClassOf` axioms the subclass appears before the superclass:
```xml

  <xsd:complexType name="SubClassOf">
    <xsd:complexContent>
      <xsd:extension base="owl:ClassAxiom">
        <xsd:sequence>
          <xsd:group ref="owl:ClassExpression"/>
          <!-- This is the subexpression -->
          <xsd:group ref="owl:ClassExpression"/>
          <!-- This is the superexpression -->
        </xsd:sequence>
      </xsd:extension>
    </xsd:complexContent>
  </xsd:complexType>
  <xsd:element name="SubClassOf" type="owl:SubClassOf"/>
```

The [current XML parser](https://github.com/phillord/horned-owl/blob/master/src/io/reader.rs#L390-L395) however will deserialize it in the opposite order:
```rust
        b"SubClassOf" => {
            SubClassOf {
                super_class:from_start(r, e)?,
                sub_class:from_next(r)?,
            }.into()
        }
```

And the [current XML serializer](https://github.com/phillord/horned-owl/blob/master/src/io/writer.rs#L698-L702) will also do it the wrong way:
```rust
contents!{
    SubClassOf, self,
    (&self.super_class,
     &self.sub_class)
}
```

The same bug also affects `SubObjectPropertyOf` and `SubDataPropertyOf`.

althonos:
Fixes #9.

### DIFF
--- a/src/io/reader.rs
+++ b/src/io/reader.rs
@@ -389,8 +389,8 @@ fn axiom_from_start<R: BufRead>(
         b"Declaration" => declaration(from_start(r, e)?),
         b"SubClassOf" => {
             SubClassOf {
-                super_class:from_start(r, e)?,
-                sub_class:from_next(r)?,
+                sub_class:from_start(r, e)?,
+                super_class:from_next(r)?,
             }.into()
         }
         b"EquivalentClasses" => {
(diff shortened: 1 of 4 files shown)


phillord:
Can you squash these two then I think we good to go?

I've added you onto the repo with push. If you have major changes would still be good to do them via PR. I'll do the same with anything major that I want to do.

PR #28, opened 2021-12-02
Bumps "quick_xml" crate to latest version.

filippodebortoli:
Adding the "encoding" feature ensures that bumping to latest version does not break compilation.


phillord:
Hmmm, this works nicely on main, but breaks on devel, giving me errors from rio for reasons that I do not immediately understand.

phillord:
Am closing this now` as I have addressed the issue in a different way.

Thanks for the PR.

PR #29, opened 2021-12-03
Introduces error handling for ParserOutput.

filippodebortoli:
Introduced enum ParserError, for now handling only cases where an unsupported extension is found.
The enum is added to the 'error' module.
Routines in 'command' have been updated to make use of the error.


phillord:
Can you check https://github.com/phillord/horned-owl/pull/36 for me to see if it does what you expect?

phillord:
Remember why I hadn't taken this now -- because I moved from failure to this_error.

PR #30, opened 2021-12-03
Refactoring command-line tools into multiplexed app

filippodebortoli:
See integration of horned-big into horned-owl.
The binaries can be compiled separately, so that horned-big can still be generated as a standalone tool.
If this is desired, I could proceed to do it for all tools and then complete the PR.


phillord:
Indeed, this is something that I would like to see done, but haven't got around to. My inclination would be to call the "main" command `horned` rather than `horned-owl` just because its shorter. With this commit, I don't understand why you have move the business logic from `horned-dump` into `horned-owl`? Why not leave it in `horned-dump`? Or this just because you are in progress?

filippodebortoli:
> Indeed, this is something that I would like to see done, but haven't got around to. My inclination would be to call the "main" command `horned` rather than `horned-owl` just because its shorter. With this commit, I don't understand why you have move the business logic from `horned-dump` into `horned-owl`? Why not leave it in `horned-dump`? Or this just because you are in progress?

The `matcher_dump` is a leftover from my previous attempt at setting up the subcommand. Should I squash together all commits when finished the refactoring?

phillord:
Yes, I think a squash would help. I'd like to call the command `horned` also.

I'm afraid I've had to touch quite a lot of the files as a result of the removal of `failure` and general error refactoring. I'd be happy to take this as PR to devel. I'm trying out development on a clean head on main which is a new one for me. I am not sure it is worth the effort.

filippodebortoli:
> Yes, I think a squash would help.
> 
> I'm afraid I've had to touch quite a lot of the files as a result of the removal of `failure` and general error refactoring. I'd be happy to take this as PR to devel. I'm trying out development on a clean head on main which is a new one for me. I am not sure it is worth the effort.
> 
> Please feel free to put yourself as author of the `horned` command!

Now all changes appear in a unique commit.
I put myself as the author of the "horned" meta-command. :-)
I believe the merge with devel might go smoothly. The only conflicts that are currently signalled are about file renaming (e.g. "src/bin/horned-big.rs", which is now "src/bin/horned_big.rs").

phillord:
I have merged this now. There were quite a few conflicts mostly because of the error handling changes I have made, so do let me know if I have broken anything!

Thanks for the code!

PR #32, opened 2022-01-17
Fix rendering of `AnnotationAssertion` subjects in OWL/XML writer

althonos:
Hi Phil, it's been a while!

While working on the `fastobo-owl` crate (which I'll release soon), I noticed the following bug with the XML writer.

Currently, it will write a `NamedIndividual` element when serializing an `AnnotationAssertion` axiom:
```xml
    <AnnotationAssertion>
        <AnnotationProperty IRI="http://www.geneontology.org/formats/oboInOwl#id"/>
        <NamedIndividual IRI="http://purl.obolibrary.org/obo/MS_0000000"/>
        <Literal>MS:0000000</Literal>
    </AnnotationAssertion>
```

However according to the [OWL2 specs](https://www.w3.org/TR/owl2-syntax/#Annotation_Assertion) the IRI must be written directly, and the file above will fail to be processed by the OWL API parser. The correct element should be:
```xml
    <AnnotationAssertion>
        <AnnotationProperty IRI="http://www.geneontology.org/formats/oboInOwl#id"/>
        <IRI>http://purl.obolibrary.org/obo/MS_0000000</IRI>
        <Literal>MS:0000000</Literal>
    </AnnotationAssertion>
```

I fixed the parser to render correctly, but I'm wondering if the `subject` field of `AnnotationAssertion` axiom shouldn't be it's own type (with the same variants as a `NamedIndividual`) to avoid future confusion like this.

### DIFF
--- a/src/io/owx/writer.rs
+++ b/src/io/owx/writer.rs
@@ -642,11 +642,21 @@ contents! {
     ClassAssertion, self, (&self.ce, &self.i)
 }
 
-contents! {
-    AnnotationAssertion, self,
-    (&self.ann.ap,
-     &self.subject,
-     &self.ann.av)
+render! {
+    AnnotationAssertion, self, w, m,
+    {
+        (&self.ann.ap).render(w, m)?;
+        if let Individual::Named(named) = &self.subject {
+            // render in an <IRI> or <abbreviatedIRI> tag directly
+            named.0.render(w, m)?;
+        } else {
+            // render in an <AnonymousIndividual> tag
+            self.subject.render(w, m)?;
+        }
+        (&self.ann.av).render(w, m)?;
+
+        Ok(())
+    }
 }
 
 render! {


phillord:
I think that the data model is probably broken. In fact, the struct says:

```rust 
    // Annotation Axioms
    /// An annotation assertion axiom
    ///
    /// States that `annotation` applies to the
    /// `annotation_subject`. Annotations refer to an `IRI` rather
    /// than the `NamedEntity` identified by that `IRI`.
    AnnotationAssertion {
        subject: Individual,
        ann: Annotation
    },
```

So the documentation is correct but the code is wrong. This is a regression actually. It used to be:

```rust
 AnnotationAssertion {
        subject:IRI,
        ann: Annotation
    },
```
This is correct but incomplete and I changed it when I was introducing `AnonymousIndividual` which are required according to the structural spec.

```
AnnotationSubject := IRI | AnonymousIndividual
```

So, the subject should be `IRI|AnonymousIndividual` which I don't think I have an enum for at the moment. Once I do that, I am not sure whether I need the `Individual` enum any more. And I guess, I need to check everywhere that `AnonymousIndividual` should be, because I've just noticed that the `AnnotationValue` enum is wrong because it's missing `AnonymousIndividual`.

OWL is a bit complicated, apologies!

althonos:
No worries! I also think that there should be a dedicated `AnnotationSubject` here.

By the way, regarding the model, is there a reason for having separate declaration axioms (`DeclareClass`, etc.) while the spec has only one that wraps an `Entity`? I think this may end up introducing some bugs when serializing because you'd need to define special cases for these (which is fine, and works in my `horned-functional` code currently, but I was just wondering if there's a specific reason). Cheers!

phillord:
Okay, I think we go for the `AnnotationSubject` as the right option. I suspect `Individual` can die at that point.

Why multiple declaration axioms? Honestly I don't remember, but probably because it makes things like `AxiomMappedOntology` work faster, because I can simply check for all instances of `DeclareClass` rather than checking all instances of `Declare(NamedEntity)`. However, now that I can have multiple indexes and `Ontology` is abstract, that requirement has gone.

So, it should probably go. However, it is working at the moment, so I think this is something that should go into a future release. The `devel` branch has probably reached a point where it has enough changes now!

althonos:
> Okay, I think we go for the AnnotationSubject as the right option. 
Then I'll close this PR, the XML serialization issue will be fixed by the new object model.

> So, it should probably go. However, it is working at the moment, so I think this is something that should go into a future release. The devel branch has probably reached a point where it has enough changes now!
Yes I agree, I was just wondering. It makes total sense with an `AxiomMappedOntology`.

phillord:
Going to re-open this or I will forget:-)

phillord:
Closed in favour of https://github.com/phillord/horned-owl/commit/ed4a4b72231408a06f74fa4fff81638aad58465a

PR #33, opened 2022-01-19
Improve `Builder` interface to allow skipping copy when possible

althonos:
Hi again, this time with a smaller PR to improve the way `Build` creates IRI strings.

In the current version of the code, `Build::iri` would always copy the string passed as argument to a new buffer, even when the cache was hit, because it needed an `IRI` struct to check for membership in the cache. Which means that every call to `Build::iri` was copying the string contents.

By implementing `Borrow<str>` for IRI, you can now check whether the IRI is already in the cache simply by comparing to its string value. This means `Build::iri` (and associated methods) can be updated to take any `AsRef<str>` implementor.

### DIFF
--- a/src/io/owx/reader.rs
+++ b/src/io/owx/reader.rs
@@ -171,18 +171,37 @@ fn read_event<R: BufRead>(read: &mut Read<R>) -> Result<(Vec<u8>, Event<'static>
     }
 }
 
-fn decode_expand_curie_maybe<R: BufRead>(r: &mut Read<R>, val: &[u8]) -> Result<String,ReadError> {
-    let s = r.reader.decode(val)?.to_string();
-    Ok(expand_curie_maybe(r, s))
+fn decode_expand_curie_maybe<'a, R: BufRead>(r: &mut Read<R>, val: &'a [u8]) -> Result<Cow<'a, str>, ReadError> {
+    // Okay, so a lot of matching, but without this the borrow checker
+    // is gonna complain. This let's us do the following:
+    // - if the CURIE can be decoded without replacement, and if it is
+    //   already a complete IRI, we don't have to copy it, and we can
+    //   return it as as borrowed string.
+    // - in any other case, we need to perform a copy, otherwise the decoded
+    //   string / unabbreviated IRI is not going to live long enough.
+    #[cfg(feature = "quick-xml/encoding")]
+    match r.reader.decode(val) {
+        Cow::Borrowed(b) => expand_curie_maybe(r, b),
+        Cow::Owned(o) => match expand_curie_maybe(r, &o) {
+            Cow::Borrowed(b) => Cow::Owned(b.to_string()),
+            Cow::Owned(o) => Cow::Owned(o),
+        },
+    }
+
+    #[cfg(not(feature = "quick-xml/encoding"))]
+    match r.reader.decode(val) {
+        Ok(curie) => Ok(expand_curie_maybe(r, curie)),
+        Err(e) => Err(ReadError::from(e))
+    }
 }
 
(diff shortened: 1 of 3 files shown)


phillord:
I am probably being thick, but does this require `Borrow<str>`. We already implement `AsRef<str>` for IRI? The `IRI` implementation is broken anyway, because it is `Rc` and long term I want to replace it with `IRI<R> where R:AsRef<str>`, so that we can support `Arc` cleanly.

Also, sorry to be a PITA, but I've been trying out using a `devel` branch and having a clean `main`. I'm not convinced it's works well, but hey. Can you rebase against `devel`?

phillord:
Ah, yes, because the underlying `BTreeSet` requires `Borrow` because of it's guarantees around hashing. Hmmm.

althonos:
Exactly! You could have `IRI<R> where R: Borrow<str>` to make sure that the hash values remain consistent for the storage, but then use `AsRef<str>` for the input (you need to temporarily take a reference to a string, but don't need as much restriction as `Borrow`). `Borrow<str>` is implemented for `Rc<str>` and `Arc<str>` as well.

I'll rebase against `devel` then :+1:

phillord:
The last question, then, is whether the increased strictness of the `Borrow<str>` is a problem, especially as if we are introducing it only because of `Build` which is not an essential part of the model interface.

I think you raise a good option with having `Build` require `Borrow<IRI>` but `IRI` remaining `AsRef<str>`. As you say, it makes no difference for `Rc` or `Arc`, but it is a significantly stricter interface otherwise.

phillord:
Oh, out of curiosity, have you profiled these changes against any significant load? I've got no problems with adding them regardless, because it's always good to be neat.

althonos:
> The last question, then, is whether the increased strictness of the Borrow<str> is a problem, especially as if we are introducing it only because of Build which is not an essential part of the model interface. 

I think it's fair to require `Borrow<str>` for IRI because of the hash immutability guarantee. In `Build` methods you'd only require `AsRef` so that would make it functionally equivalent to what it is right now. Anyway, `String`, `&str`, `Arc<str>`, `Rc<str>`, `Cow<str>` all implement `Borrow<str>` so it's not going to be utterly restrictive to the end user.

> Oh, out of curiosity, have you profiled these changes against any significant load? I've got no problems with adding them regardless, because it's always good to be neat.

I saw no difference with the `io_read` benchmark but since it contains absolutely no duplicate IRIs it's going to be the worst case. I'll make a benchmark with a document that contains redundant IRIs.

althonos:
On a real-life example (an OWX version of the [PSI MS controlled vocabulary](https://github.com/HUPO-PSI/psi-ms-CV/)) I get the following results:
```
 name     main.bench ns/iter     iri-ref.bench ns/iter  diff ns/iter  diff %  speedup 
 ms_read  155,119,352 (37 MB/s)  143,849,698 (40 MB/s)   -11,269,654  -7.27%   x 1.08
```

phillord:
My feeling is I would still like to put the minimum requirements on `IRI` possible. I don't think we need `Borrow` on `IRI` just as a requirement on `Build`. Obviously you have done the majority of the work on this, so I'm happy to try this; alternatively if you are happy to give it a go, it might happen quicker.

A 10% speed up from this is pretty good, I think!

althonos:
I think that the way IRI are managed is going to change nevertheless when you actually do the refactor, so you either 1. merge this PR, and think about removing the `Borrow` requirement later when you actually implement the IRI refactor, or 2. wait for the IRI refactor, and then I'll update my code. But with the current API there's no issue implementing `Borrow<str>` for `IRI` and as a consumer I'd take some performance improvement any time :wink:

PR #34, opened 2022-02-01
Fix/annotation render regression

phillord:


phillord:
@althonos How does this look?

althonos:
LGTM! I have a little follow-up patch regarding `horned_owl::io::owx::writer::write` that I'd also like to see in the `0.10.1` release, if you can wait until then :+1:

phillord:
That's fine. Bung a patch in. There are a few other changes. 

It will be 0.11, I think. I might start doing point releases when I manage to pay someone to work on this!

PR #35, opened 2022-02-02
Update `horned_owl::io::owx::writer::write` to use a generic argument

althonos:
This just changes the main OWX writer function to support a generic argument. Current signature requires a `&mut dyn Write` argument, which 1. uses a `dyn` object which has a runtime cost, and 2. cannot take values by ownership. Now you can do the following (while currently you'd need a mutable `f` and pass a mutable reference):
```rust
let f = std::fs::File::create("out.owx").unwrap();
horned_owl::io::owx::writer::write(f, ont, None); 
```

### DIFF
--- a/src/io/reader.rs
+++ b/src/io/reader.rs
@@ -59,16 +59,16 @@ where
     ns_buf: Vec<u8>,
 }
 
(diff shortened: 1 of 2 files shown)


althonos:
Great!

PR #36, opened 2022-02-03
Filippodebortoli command error management

phillord:
Conflict resolve for https://github.com/phillord/horned-owl/pull/29


althonos:
I'd think it would be more ergonomic to have a `PathBuf` rather than a `Box<Path>`, there's no advantage using the latter IIRC.

phillord:
Think that is a good call

PR #44, opened 2022-07-04
Add convenience method for axiom kind

jannahastings:


jannahastings:
Yes sure! That sounds sensible

jannahastings:
> Want to push the changes up as a new commit? The rename will hit quite a few places am afraid.

Have a look? Seems to be just the OWX reader and writer, as far as I can tell? (and py-horned-owl but I will update that separately)

phillord:
Looks good to me, except can you add a short note to releases.md? 

Other than that, I've added you as a collaborator, so feel free to merge to devel.

Issue #46, opened 2022-10-11
Missing namespaces from OWX round trip


phillord-ncl:
https://github.com/jannahastings/py-horned-owl/issues/6

phillord-ncl:
The OWL API produces this:

```
<Ontology xmlns="http://www.w3.org/2002/07/owl#"
     xml:base="http://www.example.com/iri"
     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
     xmlns:xml="http://www.w3.org/XML/1998/namespace"
     xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
     ontologyIRI="http://www.example.com/iri"
     versionIRI="http://www.example.com/viri">
    <Prefix name="o" IRI="http://www.example.com/iri#"/>
    <Prefix name="owl" IRI="http://www.w3.org/2002/07/owl#"/>
    <Prefix name="rdf" IRI="http://www.w3.org/1999/02/22-rdf-syntax-ns#"/>
    <Prefix name="xml" IRI="http://www.w3.org/XML/1998/namespace"/>
    <Prefix name="xsd" IRI="http://www.w3.org/2001/XMLSchema#"/>
    <Prefix name="rdfs" IRI="http://www.w3.org/2000/01/rdf-schema#"/>
...
```

Where the prefixes are both specified as namespaces and prefixes. Horned just produces the prefixes.

I may be slightly biased, but horned's approach seems more sensible. I have no ideal at all who is closer to the spec.

@ignazio1977  @jamesaoverton Anyone got any ideas who is right? (Or rather, ignazio1977 will be right, of course, but why is he right?)

ignazio1977:
I'm hardly ever right (just ask my kid), however here the difference is that the namespaces are XML functionality and the prefixes are syntax specific. So with the XML namespaces you can shorten XML elements, without you might have illegal XML, depending on the renderer.

phillord:
Thanks to @Eugenio2192, I have added an attempt at fixing this now.

I'd welcome feedback on whether this work for its intended purpose.

areleu:
I think this can be closed, the namespaces are rendering properly. 

Regarding the pretty printing, I was working on a PR in https://github.com/tafia/quick-xml/pull/731 but it may take some time until they decide if they merge it so we might need to wait for that.

phillord:
[...]

PR #47, opened 2022-10-16
Attempt to add namespaces to owx writer

Eugenio2192:
Probably a little naive, I am still  new to Rust and this whole library is a little overwhelming. Tried to do an extra Render implementation of PrefixMapping but it got weird and I gave up.

New lines and indentation within the opening tag are not yet working. I took a dive into quick_xml and my first impression is that this is not something that is there but I might be wrong.

Addresses #46


phillord:
Thanks! I will give this a go.

Part of the problem is that we are using a slightly old version of quick-xml and the XML namespace handling has all changed in the latest version (there is a `NSReader` now). So, all a bit of a mess and I have to work out what to update first.

Issue #48, opened 2022-10-19
quick-xml version is long out of date


phillord-ncl:
The name space handling is all changed in the new version of the reader so the change is non-trivial unfortunately.

filippodebortoli:
Some observations:

  + `quick_xml::reader::{NsReader,Reader}::read_text()` is only implemented for `NsReader<&'i [u8]>`. This is currently used in line 369 of `reader.rs`. Same goes for `read_to_end()`. The `Read` struct uses `Reader<R: BufRead>', which means that these functions are unavailable.
  + `read_namespaced_event()` is now superseded by `read_resolved_event_into()` ([Link](https://docs.rs/quick-xml/latest/quick_xml/reader/struct.NsReader.html#method.read_resolved_event_into)), which does not require `ns_buf` but returns a tuple `(ns, event)`
  + `e.local_name()` is not returning a `&[u8]` anymore. This can be fixed by changing it to `e.local_name().as_ref()`
  + `decode()` is now available from `r.reader.decoder().decode()`.

These seem to be the biggest issues in `reader.rs` when switching to version `0.25.0`.

filippodebortoli:
A temptative new version of `read_event`:

```rust
fn read_event<A: ForIRI, R: BufRead>(
    read: &mut Read<A, R>,
) -> Result<(Vec<u8>, Event<'static>), HornedError> {
    let r = read
        .reader
        .read_resolved_event_into(&mut read.buf)?;

    match r {
        (_, Event::Eof) => Err(invalid! {
            "Unexpected EoF at {}", read.reader.buffer_position()
        }),
        (resolution, event) => {
            match resolution {
                ResolveResult::Unbound => Ok((Vec::new(), event.into_owned())),
                ResolveResult::Bound(Namespace(ns)) => Ok((ns.to_vec(), event.into_owned())),
                ResolveResult::Unknown(_) => todo!(),
            }
        }
    }
}
```

phillord:
Completed!

PR #50, opened 2022-11-17
Bump quick-xml to latest version (0.26.0)

filippodebortoli:
Ok, here is my attempt at bumping this dependency.
I had to change some function signatures to read parts of the `Read` struct to avoid complaints from the compiler.
The biggest changes are that `Read` currently comes without buffers and `read_event` has been removed and integrated into the main reading loop.
There is room for improvement, especially considering that `quick-xml` has now newer APIs (see e.g. `try_get_attribute()`).

All tests are successful when the `encoding` feature is not activated.
I have not tried (managed) to activate the feature and complete the testing.

Addresses [#48](https://github.com/phillord/horned-owl/issues/48)


phillord:
Sorry for very slow merge! Busy period of the year.

The changes look good, and I am happier with updates that need room for improvement, over no update at all!

Thanks!

PR #51, opened 2022-11-29
Extend README with description of command line tools.

filippodebortoli:
I sketched an extension of the README where the tools associated with horned are briefly presented.
This should allow the user to get to use them without having to look at the source code of each subcommand.

I also added a few lines on how to include the library in Cargo.toml (redundant given the existence of crates.io) and shields that point to the crates.io and docs.rs pages associated to horned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,10 @@
 Horned OWL
 ==========
 
-Horned-OWL is a library for manipulating OWL (Ontology Web Language)
-data. While there are a number of libraries that manipulate this form
+![Crates.io](https://img.shields.io/crates/v/horned-owl?style=flat-square) ![docs.rs](https://img.shields.io/docsrs/horned-owl?style=flat-square)
+
+Horned-OWL is a library for manipulating data written using the [Web Ontology Language (OWL)](https://en.wikipedia.org/wiki/Web_Ontology_Language).
+While there are a number of libraries that manipulate this form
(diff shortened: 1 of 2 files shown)


phillord:
While I think this is good, I have been thinking for a while about removing the binaries to their own crate. Or to refactor the repo, probably, into a multi-crate workspace. This didn't make sense originally, but now there is a good way of using horned without using the binaries, it probably does now.

What are your thoughts on this?

althonos:
I'd be in favour of splitting the crate into a multi-crate, as right now compiling `fastobo-owl` means building `horned-owl` and compiling the code / dependencies for the binaries (such as `clap`) that are not used in the library interface :+1:

phillord:
I will am inclined then to do this. As this will be a substantial change to the code base, it will probably merit a release all by itself.

filippodebortoli:
> While I think this is good, I have been thinking for a while about removing the binaries to their own crate. Or to refactor the repo, probably, into a multi-crate workspace. This didn't make sense originally, but now there is a good way of using horned without using the binaries, it probably does now.
> 
> What are your thoughts on this?

I am also in favor of the split.
Like @althonos, I am only interested in using the library and as such, I would be happier with a lightweight dependency.
Plus, I think it generally makes sense to have a library and then separate crate(s) for the tools.

Issue #54, opened 2023-01-31
horned with no options panics


phillord:
It should print the help out

~/src/rust/horned-owl/main/horned-bin$ horned
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', horned-bin/src/bin/horned.rs:38:11
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

filippodebortoli:
I opened PR #55 that should be fixing this issue.

filippodebortoli:
Fixed by #55.

PR #55, opened 2023-02-01
Horned (binary) now defaults to help when subcommands are missing.

filippodebortoli:
In addition, "compare" is now fully integrated into the subcommands list.

Solves #54


phillord:
Should have done this a while back but sent an invite. Would you like to merge this yourself?

filippodebortoli:
Thank you for the invitation! PR is now merged.

Issue #56, opened 2023-02-02
Improving IRI handling using `oxiri`


filippodebortoli:
At the moment, `horned` comes with its own IRI data structures `IRI` and `IRIString`, plus related traits `ForIRI` and `WithIRI`, methods for IRI resolution, and so on.

I found out that the @oxigraph project provides the [oxiri](https://github.com/oxigraph/oxiri) crate, which comes with most of the features that `horned` seems to need.

@phillord do you think `horned` could benefit from using `oxiri` for handling IRIs or would you stick with the current implementation?

althonos:
This sounds good too since `horned` uses the `curie` crate ATM, which is not super actively maintained...

phillord:
Interesting. I think it is worth discussing the needs behind this first, rather than the specifics of the implementation.


Currently, `IRI` does two things -- at a trait level it is a new type around, well, mostly `Borrow<str>` with a pile of other traits; and, in implementation, it allows interning via `Rc` and `Arc`. 

This addition changes `IRI` to allow do IRI validation as well as giving easy access to all various components of the IRI. I guess this is the key point in it's favour. This makes the library a bit stricter, but improves the ergonomics of IRI manipulation. Oxiri, IIUC, caches all the validation data, so if we check for protocol or host or whatever multiple times, that computation is shared.

There are several against as well. 

It makes the ergonomics of all the `Build` methods more complex, because it requires error handling. 

It makes some lazy code useful for testing not work:

```
use horned_owl::model::*;
fn main() {
    let b = Build::new_rc();
    println!("{:?}", b.class("123"))
}
```

Similarly, it would make parsing OWL which made use of invalid IRIs impossible, since the validation is there at a type level.


We could achieve some of the advantages without the cost. So, if `model::IRI` implemented `TryInto<Iri>`, we could get all the validation and so forth, without the changed interface to `Build`. But, we'd loose the caching of IRI components and the IRI validation would be off by default.

Curie vs oxiri is a slightly separate issue. The current PR doesn't replace curie (although it probably could do). And we could replace curie with oxiri without changing model.

Have I covered everything? In general, it seems like not a bad idea. I am a bit concerned about the ergonomics, since this is a drip-drip kind of thing. On it's own it doesn't seem to bad, but it's easy to make a library harder to use bit at a time. I am not sure about whether IRI validation by design would ever be a problem; I don't know if the OWL API can output invalid IRIs, but there are bound to be some tools out there that do.

[...]

PR #57, opened 2023-02-03
IRI management with `oxiri`

filippodebortoli:
Addresses #56.
A lot of refactoring involved.
Many methods do now return Results, and fail if the parsed IRI is ill-formed. Tested with make pre-commit and pre-push (except for the ignored test, which is for the moment failing). Passed with cargo clippy a couple of times, there is still some rough spot.


phillord:
Closing in favour of https://github.com/phillord/horned-owl/commit/9935cfded9a19384118e3ffcbc63e5e615c8167e

PR #59, opened 2023-02-12
Add SWRL rules (not ready for merge)

phillord:
This branch will provide support for SWRL rules


phillord:
https://github.com/phillord/horned-owl/blob/feature/swrl-rules/branch.org

Documentation on this branch is now available. There is currently no support for SWRL rules at all, but I have changed all the interfaces and wound in OntologyID and DocIRI as iterable elements.

@filippodebortoli, @althonos, @balhoff Comments welcome!

phillord:
I think that this is nearly where I would want it to be to add SWRL rules.

I am still a little uneasy about the solution, but I think it is probably the best of the options I have. Given that this is a big change, my inclination (despite the name of the branch) is to merge it to devel now and probably do a new break everything release, without actually supporting SWRL rules.

The actual support can come later and can be merged more slowly.

filippodebortoli (approved):

phillord:
Lots of clean ups and renamings. Will merge soon unless anyone complains.

phillord:
Merged to devel. Want to implement one or two SWRL rules before I merge this to main in case it is all a horrible mistake

balhoff:
> Merged to devel. Want to implement one or two SWRL rules before I merge this to main in case it is all a horrible mistake

We have a bunch in RO: http://purl.obolibrary.org/obo/ro.owl

althonos:
@phillord : I think this is going to cause some issues because creating an `AnnotatedComponent` allows for invalid constructs, such as annotated `OntologyID` or annotated `Import` which are not allowed in the spec, and this will likely lead to issues with parsers and writing having to dynamically check for such constructs. I would suggest instead:

- create an `Annotated<T>` generic struct:
    ```rust
    struct Annotated<T, A: ForIRI> {
        inner: T,
        annotations: BTreeSet<Annotation<A>>
    }
    ```
    so that now you can have `Annotated<DLSafeRule<A>>` and `Annotated<Axiom<A>>`

- making `Component` the following:

    ```rust
    enum Component {
        OntologyID(OntologyID<A>),
        DocIRI(DocIRI<A>),
        OntologyAnnotation(OntologyAnnotation<A>),
        Import(Import<A>),
        Axiom(Annotated<Axiom<A>>),
        DLSafeRule(Annotated<A, DLSafeRule<A>>) // when they get implemented
    }
    ```

This only allows annotation on axioms and SWRL ruels, but then you can still view an `Ontology` as a collection of `Component`.

althonos:
Besides, having the metadata as a `Component` relaxes the uniqueness, so I'm not sure it's such a good idea, you'll need to make sure the `OntologyID` component is present only once....

althonos:
The other option is to just add the SWRL rules as `Axiom` variants, and use the `HigherKinded` to differentiate...

phillord:
You make a good point if not all Components cannot be annotated and likewise the uniqueness. But, I am, perhaps, not so worried about having the model being able to represent things that are not strictly valid. I am sure that this is possible in some of syntaxes also.

At the same token, this seems a bigger change to me, since the rest of the code uses the `AnnotatedAxiom(Axiom, Annotation)` logic; putting the `Annotation`s inside the `Axiom` is going to change quite a lot of things.

phillord:
https://github.com/phillord/horned-owl/issues/65

I think it also might be a good idea in getting the SWRL implmentation working, even if it involves fiddling later.

althonos:
I wasn't suggested to move the annotation inside the axiom though, only to have a wrapper inside components for things that can be annotated (axioms or rules) to keep the other components annotation free.

For the functional writer this is a big deal because I cannot use the code I have if the model allows illegal OWL constructs, as all the writer code is infallible. I can change it but again, given that the OWL actually specifies a model, I would be in favour of keeping it consistent with that. I can make a PR with what I mean to be concrete.

althonos:
I also think from a usability point of view it is nice to keep metadata (IRI/docIRI/imports) separate from the ontology contents, otherwise that means you cannot easily get the docIRI from a `SetOntology` without doing a full scan or turning it into an AxiomMappedOntology which I think is overkill.

phillord:
I can see your point, but the ability to treat at Ontology as a stream of things is a big advantage; it makes switching between different indexed ontologies trivial.

It would still be possible to build an ontology as

````
struct IdentifiedOntology {
      OntologyID, OntologyImports, OntologyAnnotation, Set<OntologyComponents>
}
````

and to fit within the "ontology as a set of components" interface, without going the whole `AxiomMappedOntogy` route.

PR #64, opened 2024-02-24
Manually implement `Default` to avoid introducing a bound on IRI types

althonos:
Hi!

While refactoring the parsers to be included in the crate, I realized that I can make them generic over the ontology types, provided they are `Default + MutableOntology`, because in both cases I create an empty ontology and then fill it incrementally with `MutableOntology::insert`.

I tried doing that but then got plenty of errors, because most ontology types implement `Default` with a trait requirement on the IRI generic type, which should not be needed, but is caused by the use of the `derive` implementation. Since for all ontology types creating a default one means initializing an empty collection, there shouldn't be a need for default IRI anywhere.

To address this I manually implemented `Default` on all index and ontology types. I also changed the type signatures of the index types to be more consistent: `OneIndexedOntology` didn't have any type boundary in the type signature while the remaining indexed ontology had one. I removed the type boundaries where they were not needed, otherwise it created a whole kind of trait dependency issues in the trait implementations.

In general I've observed that library interfaces have much more flexibility when the type boundaries are as lax as possible, and there are many places here where `ForIRI` is required where it would actually not be needed (for instance in `Clone` implementations). This would be nice to standardize :smile:

### DIFF
--- a/src/ontology/component_mapped.rs
+++ b/src/ontology/component_mapped.rs
@@ -54,7 +54,7 @@ macro_rules! onimpl {
     };
 }
 
-#[derive(Debug, Default, Eq, PartialEq)]
+#[derive(Debug, Eq, PartialEq)]
 pub struct ComponentMappedIndex<A, AA> {
     component: RefCell<BTreeMap<ComponentKind, BTreeSet<AA>>>,
     pd: PhantomData<A>,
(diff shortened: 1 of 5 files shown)


filippodebortoli (approved):

filippodebortoli:
> To address this I manually implemented `Default` on all index and ontology types. I also changed the type signatures of the index types to be more consistent: `OneIndexedOntology` didn't have any type boundary in the type signature while the remaining indexed ontology had one. I removed the type boundaries where they were not needed, otherwise it created a whole kind of trait dependency issues in the trait implementations.

Nice, I had a look at the code and there don't seem to be any issues, so I approve the change. Thanks!
Unless @phillord has something to object in the next few days, I would perform the merge myself.

> In general I've observed that library interfaces have much more flexibility when the type boundaries are as lax as possible, and there are many places here where `ForIRI` is required where it would actually not be needed (for instance in `Clone` implementations). This would be nice to standardize 😄

Agreed, if the requirements can be simplified we should go for that. 👍

althonos:
Thanks! I'll merge then :+1:

> As a general point, I would agree with the idea that it is nice in theory to keep the bounds as lax as possible. However, the counter to this, is that when I made `IRI` generic, I found that it was extremely hard to achieve. I tried adding bounds where the compiler told me I needed them; but in the end, I got into a repeated cycle of "compile, read error, add new bound"; and the method signatures all got really, really long. This is why I created the `ForIRI` trait and followed a fairly general policy of adding A: ForIRI everywhere; the alternative was too painful and, I think, largely unmaintainable.

Yes I agree, this is indeed the way to go when having several trait requirements, that's done e.g. in `petgraph` with `petgraph::graph::IndexType`, but then `petgraph` still allows you to use `Graph::new` without requiring the index to be `IndexType`, only the actual indexing methods have a bound, etc.

PR #66, opened 2024-02-27
Add the `AnonymousIndividual` variant to `AnnotationValue`

althonos:
Hi again,

While refactoring the OWL Functional parser I downloaded all the OWL ontologies from the OBO foundry, converted them to `.ofn` with `robot`, and then tried to parse them; most passed, some failed because of typos in the source files, but a large chunk of the ones that were failing were caused by `horned-owl` not supporting anonymous individuals as annotation values.

I added the `AnonymousIndividual` variant to the `AnnotationValue` enum, and updated the OWX and RDF/XML parsers and writers to take into account the new variant (with additional tests). With this, I can get the functional parser to pass on almost all ontologies, and probably merge it in here :smile:


filippodebortoli (approved):

phillord:
Good fix. Just something I missed from the specification.

For the functional parser, you might want to try updating the tests in `src/ont`. You should be to generate functional syntax for all of the existing tests; just checking whether they roundtrip would give you a fairly comprehensive test set.

PR #68, opened 2024-03-01
Add parser and serializer for the OWL2 Functional-style syntax

althonos:
This still needs some work, but I have most of the types in place. 

- The grammars are in a separate folder because if I get to add the Manchester parser as well, some of them can be shared between the two parsers.
- The writer will return with an error when trying to serialize an ontology that contains more than one `OntologyID` component, and will not serialize `DocIRI` components.

To-do list:

- [x] ~Some things need to be updated in the `curie` crate first so that I can handle the case of default prefixes properly, I will open a PR there.~
- [x] ~The round-tripping tests fail at the moment because some files contain both a prefix and a default prefix for the same IRI, I need to edit the files first~.
- [x] ~Figure out what to do with empty prefixes: the way the syntax allows CURIEs and the way the `curie` crate treat default namespace is a bit misleading so I need to figure out how to reconcile the two.~ -- Abbreviated IRIs with empty prefixes are now using `Curie(Some(""), local)` rather than `Curie(None, local)`; the default prefix is not used anywhere. 
- [ ] Make the parser support iteration rather than reading the full ontology, requires a bit more work because `pest` can only lex complete strings.


phillord:
Did you make the test files in `src/ont` via hand?

I realise that setting up tawny.owl and bubo for the purpose is a bit of a PITA (which is why I commited all the src/ont files to the repo!), but it is a good way of getting all the OWL files in different formats.

phillord:
Ah, okay, generated from the RDF format or equivalent?

My inclination would be to add owl-function to the Makefile in bubo -- not because it like my code better than Robot, but it does give a single source of truth for the OWL files. But, you will lose the hand formatting, alas.

What do you think? Happy to add support for this myself.

phillord:
Hmmm. Interestingly, some of the OWL test files do not have a bubo equivalent (for example, two-class-with-some) and are not used anywhere in code (except in your OFN parser). And there are a few manual files (family and family-other) which really should be in the "manual" directory because I have generated them in an entirely different way.

It looks like I need a bit of a clean up of the test files.

phillord:
Ah, figured it out. I remember getting bored with the `one-class.clj` naming scheme that I introduced at some point. So, we have files like `one-or.owl` and `or.owl` which are largely identical (semantically anyway). Both what have been produced by what is now `or.clj`, but are a bit different because they will have used different versions of the OWL API or tawny bubo.

What is the easiest way forward? I could add a fix to `devel` and you could pull it across to this branch? Or just fix it here?

althonos:
Yes I generated them from the OWL/XML sources, IIRC. I'm happy if you want to generate them with anything else, you can do so in `devel` and I'll pull them here (I'll have to do rebase anyway to include the changes from #67 which I think should be merged first).

phillord:
I have checked now. There are actually quite a few files in OWL that do not exist as bubo files. I think that this is a bit of a mess and needs cleaning. But I don't think you need to wait for me to do that. So, feel free to merge, and I will fix the test files after.

althonos:
Made a new branch from scratch based on the `devel`, now should compile without issue. I'll make an iterative parser when I get the time but I'd rather have this merged now to avoid another rebase hell :sweat_smile:

phillord:
Looks good.

Could you replace the test macros with the test-generator macro? It's should be quite straight-forward and will ensure that all the test files are being tested. This was why I caused you all the rebase pain -- might as well use it.

althonos:
Done, sorry for the delay :+1:

filippodebortoli (approved):

phillord:
In your tests, how are you checking to see whether the document has been parsed semantically correctly? As far as I can see the reader tests look to see whether the reader works, and the writer checks that the roundtrip is semantically equivalent to the input. But I think that any paired errors would not be picked up.

The RDF reader compares to the XML reader, while the XML reader does hard-coded checks against known semantics (so effectively the RDF reader does as well). Could we add that as well for ofn?

althonos:
I'm testing against the XML reader as well now. The only thing I'm not testing is the prefix mapping, because it seems to behave differently (but also in the `bubo` output) with regards to the `#` separator.

phillord:
I think that this is fine. The prefix mapping is a bit erratic at the moment -- not helped by the inability to get this information out of quick_xml. Feel free to merge!

I am afraid this is going to break when I finished SWRL support -- I might need some help with fixing this.

althonos:
Great, thanks! I'll have to come back to this as well to write a line-based parser when I have the time. Currently the parser can parse SWRL syntax but just ignores it, this will only need a `FromPair` implementation for the different SWRL types.

phillord:
I haven't quite sorted out all the types for SWRL yet, but I think that they are close now. Still deciding what to do with `Variable`.  Once that is done, hopefully, it should be plain sailing.

Issue #69, opened 2024-03-07
DataUnionOf handling is broken


phillord:
The file `data-union.owx` and equivalents does not contain a union but an intersection (which is tested in another file!).

Fixing this suggests that both OWX and RDF parsers appear to be broken and cannot handle this construct.

filippodebortoli:
I close the issue as it was fixed by #70 and #71.

PR #70, opened 2024-03-08
Adds missing handling of DataUnionOf in RDF and OWX readers.

filippodebortoli:
Addresses #69.

The related tests in `io::owx::reader` could be improved, the reason the mishandling could not be detected is that the test `io::owx::reader::test::datatype_union()` only checks that one type definition appears in the ontology, no matter what its kind is. The test `io::rdf::reader::test::datatype_union()` seems also not to be informative, as it would in this case just compare the two faulty files.

The writers seemed to already handle this construct properly.

### DIFF
--- a/src/io/owx/reader.rs
+++ b/src/io/owx/reader.rs
@@ -1025,6 +1025,11 @@ from_start! {
                         till_end(r, b"DataIntersectionOf")?
                     )
                 }
+                b"DataUnionOf" => {
+                    DataRange::DataUnionOf(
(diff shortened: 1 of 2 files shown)


phillord:
Oh, I wasn't expecting you to fix that! I've done (the same) fix on another branch, but it was just to remind me.

Happy for your to merge this. I agree about the tests; they are not comprehensively in testing the right thing. I think we could do with a few more files like `family` to test a bit more comprehensively.

PR #73, opened 2024-03-13
Parametrize all IO tests

phillord:
I have started work on paramertisizing all the roundtrip tests (which should be all the tests in XML writer, RDF reader and writer and functional when it arrives).

As expected, I have found one bug already from a missing test.


phillord:
Currently blocked on https://github.com/phillord/pretty_rdf/issues/6

phillord:
All ready to go. Any new file added to the bubo directory now will be tested for XML round tripping, comparison between XML and RDF, and RDF roundtripping. XML reading needs to be testing explicitly (although we also read XML for many of the other tests) and is where the majority of hand-written checks should go.

For functional and any other format that we choose, we can just replicate the XML -> XXX and XXX roundtripping comparison.

Cargo.toml
```diff
@@ -29,10 +29,12 @@ ureq={version="2.1.1", optional=true}
 
 [workspace]
 members=["horned-bin"]
+default-members=[".", "horned-bin"]
 
 [workspace.dependencies]
 indexmap="1.0.2"
-pretty_rdf="0.2.0"
+##pretty_rdf={path="./pretty_rdf"}
```
filippodebortoli, line 36: Could this line be eliminated?

src/io/rdf/writer.rs
```diff
@@ -33,16 +33,27 @@ pub fn write<A: ForIRI, AA: ForIndex<A>, W: Write>(
...
+        let mut f = PrettyRdfXmlFormatter::new(write, ChunkedRdfXmlFormatterConfig::all().prefix(p))?;
+        ont.render(&mut f, &mut bng)?;
+        f.finish()?;
+    }
+
+
 
     // for i in f.triples() {
     //     println!("{}", i.printable());
     // }
```
filippodebortoli, line 55: Will we ever need this debugging loop again, or can we delete it?

horned-bin/src/bin/horned_triples.rs
```diff
@@ -107,9 +109,9 @@ pub(crate) fn matcher(matches: &ArgMatches) -> Result<(), HornedError> {
 
         let mut f: pretty_rdf::PrettyRdfXmlFormatter<String, _> =
             pretty_rdf::PrettyRdfXmlFormatter::new(
-                b,
-                pretty_rdf::ChunkedRdfXmlFormatterConfig::all(),
-            )?;
+                 b,
+                 pretty_rdf::ChunkedRdfXmlFormatterConfig::all(),
+             )?;
         //let mut f = rio_xml::RdfXmlFormatter::with_indentation(&b, 4)?;
```
filippodebortoli, line 115: Can this line be removed, or does it still serve a purpose for future work?

filippodebortoli (approved):
Approved. I pointed out few sections of commented code that looked like debugging leftovers to be removed.

phillord:
Yes, I will merge. I am inclined to leave all of those debugging lines in, as well as this which is even worse:

https://github.com/phillord/horned-owl/blob/9c1048bc12618cb46015befaa6ee8295d5f83fc8/src/io/rdf/writer.rs#L40-L49

At same point the writer should be configurable. It should also be able to produce turtle, ntriples and the like via rio (or a rio wrapper).

althonos:
@phillord : There is a chance I'll just close the PR and start from a fresh rebase, I can foresee the merge errors on every test file happening and that's gonna be quicker to just rewrite the parser code in the current library :+1:

Issue #24, opened 2019-08-08
Reexport `curie` crate or `PrefixMapping` to help versioning


althonos:
Currently `horned-owl` needs the user to add the `curie` dependency manually with the exact same version as the one used by `horned-owl`, because they need `PrefixMapping` available if they use a parser or a serializer. Because updating the internal `curie` version can break the code relying on it, any `curie` update is to be treated as a breaking change, which is extremely cumbersome.

filippodebortoli:
@althonos has this issue been addressed already?
If not, would this "only" require us to `pub use curie::PrefixMapping` instead of `use curie::PrefixMapping` [here](https://github.com/phillord/horned-owl/blob/d88964b3a83082bcdf8078292f32e37a85e19203/src/io/owx/reader.rs#L1) and [here](https://github.com/phillord/horned-owl/blob/d88964b3a83082bcdf8078292f32e37a85e19203/src/io/owx/writer.rs#L1)?
Or is there more to do, e.g. use wildcards in `cargo.toml` for curie?

althonos:
Maybe add a `pub module curie`, and inside of it `pub use curie::*`, this way downstream users can use `horned_owl::curie::PrefixMapping` and `horned_owl::curie::Curie`.

filippodebortoli:
Fixed by #75.

PR #75, opened 2024-03-20
Re-exporting curie.

filippodebortoli:
Addresses #24 and applies changes proposed there.


phillord:
I never used re-exporting here, but if this works well, then I am happy with it! Nice to clear a 4 year old issue also!

Issue #65, opened 2024-02-26
Support SWRL Rules


phillord:
One of the main limitations with horned at the moment is that it does not support SWRL rules. While not strictly part of OWL2, a number of ontologies do use them. Even if only a single rule is used in a large ontology, this prevents horned from handling it at all.

I am implementing this on:

https://github.com/phillord/horned-owl/tree/feature/swrl-rule-impl

phillord:
Updated the model. Think it is closer to what is needed now, although there are still some bits of the OWL API that I do not understand.

Fingers crossed it will all make sense once I start reading things in!

phillord:
I believe that this is actually complete now, although I will need to do a little testing against real ontologies.

@althonos We will need to do something about the functional parser. After testing, maybe we can merge devel to the swrl_rules branch, extend the functional parser there, ready for the merge back. What do you think?

This has been a lot of work!

phillord:
https://github.com/phillord/horned-owl/pull/84 Functional support now in progress

althonos:
@phillord Sounds good, hopefully not too much work since the grammar already supports them!

phillord:
If you have the time, I'd be happy for you to do it:-)

I've started going through your code, but inevitably it's going to take me a while to work it all out.

phillord:
Closed with 5211a12d2d197818f5697d358831009e3bd98608

PR #83, opened 2024-04-17
DRAFT: SWRL Rule Implementation

phillord:


phillord:
https://github.com/phillord/horned-owl/issues/65

phillord:
Now supporting Rule in both XML and RDF, but only with ClassAtoms

PR #85, opened 2024-05-09
Feature/swrl rule with functional

althonos:
This is based on #84, I still need to fix a few bugs and implement the serializer.

### DIFF
--- a/src/io/ofn/reader/from_pair.rs
+++ b/src/io/ofn/reader/from_pair.rs
@@ -18,6 +18,8 @@ use super::Rule;
 
 // ---------------------------------------------------------------------------
 
+
+
 type Result<T> = std::result::Result<T, HornedError>;
 
 /// A trait for OWL elements that can be obtained from OWL Functional tokens.
@@ -689,11 +691,10 @@ impl<A: ForIRI> FromPair<A> for ClassExpression<A> {
                 let dp = DataProperty::from_pair(pair.next().unwrap(), ctx)?;
                 let next = pair.next().unwrap();
                 if next.as_rule() == Rule::DataProperty {
-                    unimplemented!() // FIXME!!!
-                                     // Err(Error::custom(
-                                     //     "cannot use data property chaining in `DataSomeValuesFrom`",
-                                     //     next.as_span(),
-                                     // ))
+                    Err(HornedError::invalid_at(
+                        "horned-owl does not support data property chaining in `DataSomeValuesFrom`",
+                        next.as_span(),
+                    ))
                 } else {
                     let dr = DataRange::from_pair(next, ctx)?;
                     Ok(ClassExpression::DataSomeValuesFrom { dp, dr })
@@ -704,11 +705,10 @@ impl<A: ForIRI> FromPair<A> for ClassExpression<A> {
                 let dp = DataProperty::from_pair(pair.next().unwrap(), ctx)?;
                 let next = pair.next().unwrap();
                 if next.as_rule() == Rule::DataProperty {
-                    unimplemented!() // FIXME!!!
-                                     // Err(Error::custom(
-                                     //     "cannot use data property chaining in `DataAllValuesFrom`",
-                                     //     next.as_span(),
-                                     // ))
+                    Err(HornedError::invalid_at(
+                        "horned-owl does not support data property chaining in `DataAllValuesFrom`",
+                        next.as_span(),
+                    ))
                 } else {
                     let dr = DataRange::from_pair(next, ctx)?;
                     Ok(ClassExpression::DataAllValuesFrom { dp, dr })
@@ -983,9 +983,6 @@ macro_rules! impl_ontology {
                 for pair in pairs.next().unwrap().into_inner() {
                     let inner = pair.into_inner().next().unwrap();
                     match inner.as_rule() {
-                        // FIXME: SWRL rules are not supported for now
-                        // Rule::DGAxiom => unimplemented!(),
-                        // Rule::Rule => unimplemented!(),
                         Rule::Axiom => {
                             ontology.insert(AnnotatedComponent::from_pair(inner, ctx)?);
                         }
--- a/src/io/ofn/reader/from_pair.rs
+++ b/src/io/ofn/reader/from_pair.rs
@@ -426,14 +426,12 @@ impl<A: ForIRI> FromPair<A> for AnnotatedComponent<A> {
             Rule::DLSafeRule => {
                 let mut inner = pair.into_inner();
                 let annotations = FromPair::from_pair(inner.next().unwrap(), ctx)?;
-                let mut head = Vec::new();
-                for pair in inner.next().unwrap().into_inner() {
-                    head.push(FromPair::from_pair(pair, ctx)?);
-                }
-                let body = Vec::new();
-                for pair in inner.next().unwrap().into_inner() {
-                    head.push(FromPair::from_pair(pair, ctx)?);
-                }
+                let body = inner.next().unwrap().into_inner()
+                    .map(|pair| FromPair::from_pair(pair, ctx))
+                    .collect::<Result<Vec<_>>>()?;
+                let head = inner.next().unwrap().into_inner()
+                    .map(|pair| FromPair::from_pair(pair, ctx))
+                    .collect::<Result<Vec<_>>>()?;
                 Ok(Self::new(
                     crate::model::Rule::new(head, body),
                     annotations
--- a/src/io/ofn/reader/from_pair.rs
+++ b/src/io/ofn/reader/from_pair.rs
@@ -543,9 +543,9 @@ impl<A: ForIRI> FromPair<A> for Atom<A> {
             Rule::AtomDataProperty => {
                 let mut pairs = inner.into_inner();
                 let pred = FromPair::from_pair(pairs.next().unwrap(), ctx)?;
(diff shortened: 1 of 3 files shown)


phillord-ncl:
Looks good so far.

Can you PR it against the feature/swrl_rule_with_function branch rather than devel? While I was playing with your code, I've started to fix a few other things (basically, updating horned-bin to work with ofn); I can add those after you've finished.

Then we can merge to everything to devel and take stock of what, if anything, needs to be done before 1.0.

althonos:
I'm done implementing the writer as well.

The one strange thing is that the order or the body and head atoms in the functional files seems to be always reversed from the XML files; I made the parser and the serializer reverse the atom vectors when reading/writing but that seems a bit strange, I suppose that's an implementation detail from `bubo`.

filippodebortoli (approved):

phillord:
Not sure I understand the "reversed" bit.

```
 <DLSafeRule>
        <Body>
            <ClassAtom>
                <Class IRI="#A"/>
                <Variable IRI="#x"/>
            </ClassAtom>
        </Body>
        <Head>
            <ClassAtom>
                <Class IRI="#B"/>
                <Variable IRI="#x"/>
            </ClassAtom>
        </Head>
    </DLSafeRule>
```

in XML while functional:

```
DLSafeRule(Body(ClassAtom(o:A Variable(o:x)))Head(ClassAtom(o:B Variable(o:x))))
)
```

If I had been writing the spec, I would have done "head" then "body", if that is what you mean.

src/io/ofn/reader/from_pair.rs
```diff
@@ -876,7 +984,8 @@ macro_rules! impl_ontology {
                     let inner = pair.into_inner().next().unwrap();
                     match inner.as_rule() {
                         // FIXME: SWRL rules are not supported for now
-                        Rule::Rule | Rule::DGAxiom => (),
+                        // Rule::DGAxiom => unimplemented!(),
```
phillord, line 987: Can you remove the comments!

src/io/ofn/writer/as_functional.rs
```diff
@@ -434,14 +440,82 @@ impl<A: ForIRI> AsFunctional<A> for AnnotationValue<A> {}
...
     }
 }
 
 impl<A: ForIRI> AsFunctional<A> for AnonymousIndividual<A> {}
 
 // ---------------------------------------------------------------------------
 
+impl<'a, A: ForIRI> Display for Functional<'a, Atom<A>, A> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
+        use Atom::*;
```
phillord, line 453: I didn't know you could do that! Learn something every day.

althonos:
> Not sure I understand the "reversed" bit.

This can be seen in the `swrl_two_variables` files. In XML, you have in the `A1` then `A` in the body, and `B1` then `B` in the head:
```xml
    <DLSafeRule>
        <Body>
            <ClassAtom>
                <Class IRI="#A1"/>
                <Variable IRI="#x"/>
            </ClassAtom>
            <ClassAtom>
                <Class IRI="#A"/>
                <Variable IRI="#x"/>
            </ClassAtom>
        </Body>
        <Head>
            <ClassAtom>
                <Class IRI="#B1"/>
                <Variable IRI="#x"/>
            </ClassAtom>
            <ClassAtom>
                <Class IRI="#B"/>
                <Variable IRI="#x"/>
            </ClassAtom>
        </Head>
    </DLSafeRule>
```
 whereas in functional you have `A` then `A1` in body, and `B` then `B1` in head:
```
DLSafeRule(
    Body(ClassAtom(o:A Variable(o:x)) ClassAtom(o:A1 Variable(o:x)))
    Head(ClassAtom(o:B Variable(o:x)) ClassAtom(o:B1 Variable(o:x)))
)
```

phillord-ncl:
Ah, okay, I see the point. Both head and body are sets in the OWL API, so the ordering isn't defined. But I would have expected it to be consistent because the ordering of a `Set` in Java usually is and the XML and functional files are produced from exactly the same objects in memory.

We could have a poke in the OWL API to see why this happens. Be curious to see what robot does also.

Does this cause problems in testing? I had to add a normalisation step to the OWX->RDF comparison to cope with sets.

althonos:
> Both head and body are sets in the OWL API, so the ordering isn't defined. 

I assumed that was the case.

> Does this cause problems in testing?

Not really, since this only appears in one test case, and as I said I just had to reverse the vectors, but if those are sets this may not be consistent if more test cases are added. Maybe would be a good idea to check the OWL API, perhaps this should be dumped in lexicographic order instead?

phillord:
Indeed. Perhaps @ignazio1977 knows how the ordering works.

ignazio1977:
Depends on the exact version in use. It used to sort the sets, was changed to a linked set on user request, so order is the order the terms were first created in.

phillord:
Confusing. As far as I can see the relevant code is in SimpleRenderer for OFN.

```
    @Override
    public void visit(SWRLRule rule) {
        sb.append("DLSafeRule(");
        writeAnnotations(rule);
        sb.append("Body(");
        render(rule.getBody(), false);
        sb.append(") Head(");
        render(rule.getHead(), false);
        sb.append("))");
    }
```

the "false" in the render call means, "don't sort". While for OWX we have in     OWLXMLObjectRenderer

```
    @Override
    public void visit(SWRLRule rule) {
        writer.writeStartElement(DL_SAFE_RULE);
        writeAnnotations(rule);
        writer.writeStartElement(BODY);
        for (SWRLAtom atom : rule.getBody()) {
            atom.accept(this);
        }
        writer.writeEndElement();
        writer.writeStartElement(HEAD);
        for (SWRLAtom atom : rule.getHead()) {
            atom.accept(this);
        }
        writer.writeEndElement();
        writer.writeEndElement();
    }
```


So, both use the iteration order of the set. This is for version4 which bubo uses. For version5 it is the stream order.

I have no idea why the two are coming out differently. It also means that we are dependent `Set` implementation ordering at best. So I don't think we can do anything to be consistent with the OWL API here that is guaranteed, and the OWL API doesn't give us guarantees.

phillord:
> Depends on the exact version in use. It used to sort the sets, was changed to a linked set on user request, so order is the order the terms were first created in.

Ah, okay. But why it is inconsistent between the two formats? Or have I been unlucky in generated XML with one version (4.5.26 according to the comments) and OFN with another (version unknown as there are no comments).

ignazio1977:
It shouldn't be inconsistent.

The order can be decided by the renderer, but the default is the order coming from the implementation of getHead/getBody, which are backed with lists, if memory serves.

Behaviour has changed in patch versions as well. It's meant to be the same in the latest 4.x and 5.x.

phillord:
And,yet, it appears to be. I've just regenerated the files for swrl_two_variables.clj and still get the same thing -- the XML output gives the body axioms in the opposite order to the functional syntax.

RDF and ttl output seem both to be in the same order as XML.

I've tested with using Robot to convert from RDF to XML and OFN. Same thing.

phillord:
I have tried lots of experiments here. My conclusion is that I am very confused.

This ontology

```
DLSafeRule(Body(ClassAtom(o:A Variable(o:x)) ClassAtom(o:A1 Variable(o:x)))Head(ClassAtom(o:B Variable(o:x)) ClassAtom(o:B1 Variable(o:x))))
```

renders as this 

```
            <ClassAtom>
                <Class IRI="#A1"/>
                <Variable IRI="#x"/>
            </ClassAtom>
            <ClassAtom>
                <Class IRI="#A"/>
                <Variable IRI="#x"/>
            </ClassAtom>
```

So the class atoms are the other way around.

I have tried with it out with a variety of ontologies with different class axioms. So this:

```
DLSafeRule(Body(ClassAtom(o:Z1 Variable(o:x)) ClassAtom(o:Y2 Variable(o:x)) ClassAtom(o:X3 Variable(o:x)) ClassAtom(o:W4 Variable(o:x)))Head(ClassAtom(o:B Variable(o:x)) ClassAtom(o:B1 Variable(o:x))))
```

renders as 

```
            <ClassAtom>
                <Class IRI="#Z1"/>
                <Variable IRI="#x"/>
            </ClassAtom>
            <ClassAtom>
                <Class IRI="#Y2"/>
                <Variable IRI="#x"/>
            </ClassAtom>
            <ClassAtom>
                <Class IRI="#X3"/>
                <Variable IRI="#x"/>
            </ClassAtom>
            <ClassAtom>
                <Class IRI="#W4"/>
                <Variable IRI="#x"/>
            </ClassAtom>

```

with the atoms in the same order. 

So I have no clear idea what could be happening. The only thing that comes to mind is that `Set` instance I am passing into the OWL API might be used directly by OWL and that this set might have a non-stable iteration order. In which case,  it is probably a Clojure oddity rather than OWL API.

phillord:
And that appears to be it. If I pass a `java.util.TreeSet` rather than a `clojure.lang.PersistantSet` into  `getSWRLRule` then the problem goes away and we get a stable order between the two. How entirely unexpected!

phillord:
I guess as we have a rough idea of the problem, but we don't know if it is going to turn into an actual problem with tests, we should merge this in now. @althonos are you happy with this?

althonos:
@phillord: If you start having more problems with testing later we can always sort the body/head in both parsers, but since for now it only causes a problem with one example (where we're lucky I guess) we can probably merge this :+1:

phillord:
All done!

PR #88, opened 2024-05-15
Tightening trait bounds

### LINKED ISSUE #74: Build should not use ForIRI constraint
There seems no reason that Build constraits the generic type with ForIRI, as none of the other structs do. It should probably be removed.

### LINKED ISSUE #87: Trait bounds should be tightened
This elaborates on #74.

The library currently contains two traits, [`indexed::ForIndex`](https://github.com/phillord/horned-owl/blob/7040db376eef3fdfeb1d6b8e240de1f1fbf27390/src/ontology/indexed.rs#L29) and [`model::ForIRI`](https://github.com/phillord/horned-owl/blob/7040db376eef3fdfeb1d6b8e240de1f1fbf27390/src/model.rs#L123), whose job is to collect trait bounds for types that can be used to build ontology indexes and IRIs.

Currently, many methods require e.g. `A: ForIRI` where it is not strictly necessary.
For instance, `model::Build::new()` should work without any bound on `A`, going from the current implementation to

```rust
pub struct Build<A>(
    RefCell<BTreeSet<IRI<A>>>,
    RefCell<BTreeSet<AnonymousIndividual<A>>>,
);

impl<A> Build<A> {
    pub fn new() -> Build<A> {
        Build(RefCell::new(BTreeSet::new()), RefCell::new(BTreeSet::new()))
    }
}
```

Another example is the implementation of `Clone` for `indexed::OneIndexedOntology` which currently is

```rust
impl<A: ForIRI, AA: ForIndex<A>, I: Clone> Clone for OneIndexedOntology<A, AA, I> {
    fn clone(&self) -> Self {
        OneIndexedOntology(
            self.0.clone(),
            self.1.clone(),
            Default::default(),
        )
    }
}
```

but may be simplified to

```rust
impl<A: Clone, AA, I: Clone> Clone for OneIndexedOntology<A, AA, I> {
    fn clone(&self) -> Self {
        OneIndexedOntology(self.0.clone(), self.1.clone(), Default::default())
    }
}
```

so that `AA` can be fully generic (which makes sense given that we use `PhantomData<AA>` there).

The goal is to remove any reference to `ForIRI` and `ForIndex` where possible, and potentially question whether these traits are necessary at all.

filippodebortoli:
Addresses #87 

Goal: minimize usage of `model::ForIRI` and `ontology::indexed::ForIndex`.

Extras:
+ [x] fixes clippy warnings for `ontology::*` ~and top-level modules~.
+ [x] adds Deref<Target = str> to ForIRI (temporarily).
+ [ ] ~ran `cargo fmt` over the codebase~ (reverted).


phillord:
We really need to unpick the cargo fmt changes which have introduced a lot of diffs. It makes it very hard to see what has been done for the trait bounds.

We should run cargo fmt before the 1.0 release, I think, and then put a check into the git hooks to standardize formatting in future. But, we need to do this in a different commit.

filippodebortoli:
> We really need to unpick the cargo fmt changes which have introduced a lot of diffs. It makes it very hard to see what has been done for the trait bounds.
> 
> We should run cargo fmt before the 1.0 release, I think, and then put a check into the git hooks to standardize formatting in future. But, we need to do this in a different commit.

Done (hopefully right), I only committed files relevant to the PR this time.

phillord:
I have updated https://github.com/phillord/horned-owl/issues/87 with my thoughts on this.

Issue #90, opened 2024-05-15
Implement additional traits for all `MutableOntology` implementors


althonos:
Consider implementing [std](https://doc.rust-lang.org/std/index.html)::[iter](https://doc.rust-lang.org/std/iter/index.html)::[Extend](https://doc.rust-lang.org/std/iter/trait.Extend.html#) and [std](https://doc.rust-lang.org/std/index.html)::[iter](https://doc.rust-lang.org/std/iter/index.html)::[FromIterator](https://doc.rust-lang.org/std/iter/trait.FromIterator.html#) for all mutable ontologies.

phillord:
Yes, this makes sense; while `FromIterator` can be simulated with `From<SetOntology>`, `Extend` cannot not be.

I will have a fiddle -- we should be able to do this with some generic cleverness over the `OntologyIndex` trait, I think.

Issue #91, opened 2024-05-15
Implement iterator marker traits for `SetOntology` iterators


althonos:
Consider implementing [std](https://doc.rust-lang.org/std/index.html)::[iter](https://doc.rust-lang.org/std/iter/index.html)::[FusedIterator](https://doc.rust-lang.org/std/iter/trait.FusedIterator.html#) and [std](https://doc.rust-lang.org/std/index.html)::[iter](https://doc.rust-lang.org/std/iter/index.html)::[ExactSizeIterator](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html#) which allows certain optimizations by the compiler.

The traits can be implemented for `horned_owl::ontology::set::SetIntoIter` and `horned_owl::ontology::set::SetIter` and perhaps some more.

PR #93, opened 2024-05-16
Support FromIterator and Extend for MutableOntology

phillord:
@althonos What about this? 

Addresses https://github.com/phillord/horned-owl/issues/90


phillord:
Also addresses #91

filippodebortoli (approved):

PR #94, opened 2024-05-16
Improving error handling for IRI resolution

filippodebortoli:
Following a `todo!()` in `resolve`, I tried to implement some basic error handling.

A few methods in the module now return `Option`s or `Result`s with minimal change to external code.

I tried to adhere to the logic of `resolve::resolve_iri()` as much as possible, though I admit I do not understand why it works this way.

### DIFF
--- a/horned-bin/src/lib.rs
+++ b/horned-bin/src/lib.rs
@@ -104,7 +104,7 @@ pub fn materialize_1<'a>(
             let local_path = Path::new(&local);
             if !local_path.exists() {
                 println!("Retrieving Ontology: {}", &i.0);
-                let imported_data = strict_resolve_iri(&i.0);
+                let imported_data = strict_resolve_iri(&i.0)?;
                 done.push(i.0.clone());
                 println!("Saving to {}", local);
                 let mut file = File::create(&local)?;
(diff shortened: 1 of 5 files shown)


phillord:
The logic in `resolve_iri` is probably a little erratic because it's a bit of the code that I knocked up quickly. I just needed enough so I could meaningfully test the RDF parser. It's also one of the bits of the code, that I think doesn't really belong in Horned at all; I was kind of hoping we could get this functionality off-the-shelf from another crate at some point, but I don't know if anyone else what written that crate yet! Still, I agree that we should stick with what we have for right now.

The PR looks mostly fine. Can you remove the big chunks of commented out code; if we ever want to look back at it, it will be in git. Have you tested it out with `horned-materialize` or any of the other binaries that user resolution?

I am happy for this to go into the 1.0 release. If we want to rewrite it wholesale or better still replace it with a third party crate, we can do that for version 2!

phillord:
@filippodebortoli 

I am on the final stages now of a 1.0 release, so if you want to merge this in please can you do so sooner rather than later!

filippodebortoli:
@phillord done. For the future, I think that `oxiri` could be a solution, since it provides IRI validation and IRI resolution methods.

phillord:
Thanks!

I've looked at oxiri and it would be nice but, unfortunately, "resolve" means "canonicalize" rather than "fetch the thing at the other end".

Issue #89, opened 2024-05-15
Standardize formatting and introduce git hooks


phillord:
The code base is currently a patchwork of formatting. 

We should standardize on the use of cargo fmt, enforced with git hooks.

filippodebortoli:
Would it be sufficient to have `.git/hooks/pre-commit` to contain the following?

```bash
#!/bin/sh

# Stash unstaged changes (optional but recommended)
git stash -q --keep-index

# Run the linter on staged files
cargo fmt # Replace with the appropriate linting command
RUSTFMT_RESULT=$?

# Unstash the stashed changes (optional but recommended)
git stash pop -q

# Exit with the linter's exit code
exit $RUSTFMT_RESULT
```

(adapted from [this](https://kinsta.com/blog/git-hooks/) blog post).

phillord:
To be honest, I think I am going to use pre-commit.com which looks nice enough.

Currently my own pre-commit checks for uncommited files and runs tests.

In terms of formatting, I would use cargo fmt -- my preference is to check only in pre-commit and break if formatting needs to happen. I think we need to consider clippy also.

The plan would be to add pre-commit first, then add each one of these in their own commits with no other substantive semantic changes. I should have done this a long time ago, but if we are going to do largescale white space, all in one is the way to go.

phillord:
This is now complete and merged. There is a "contribute.md" guide which describes how to install the hooks locally. I am running most of these in github workflow as well, so the same checks should get applied somewhere!

PR #95, opened 2024-05-17
Add basic framework for pre-commit

phillord:
This starts to address https://github.com/phillord/horned-owl/issues/89 by adding the tooling for the pre-commit framework.

This replicates the tests that I am currently using. It's all unix centric (make and sh) which can be fixed if this is a problem. I'd be happy to know, though if this works for everyone else as stands 

```
pip install pre-commit
pre-commit run --all-files
```
to get it to work.

Further tests (clippy and the like) will require substantial whitespace-code changes. So I will leave them till I can rebase against other PRs still open.


phillord:
This now adds a pre-commit framework, enforces `rustfmt` and `clippy` and applies them through out the code base.

The pre-commit is now quite heavy which is a bit of a pain, but this should result in fewer whitespace diffs in future.

Thoughts?

filippodebortoli (approved):

PR #96, opened 2024-05-30
Improving support for IRI dereferencing and comparison

filippodebortoli:
Adds what discussed recently in #87.


phillord:
I've rebased this onto devel now.

PR #97, opened 2024-05-31
Derives Ord/PartialOrd for io::rdf::reader::Term.

filippodebortoli:
Correct ordering obtained by rearranging enum variants.


phillord:
Few minor tweaks, otherwise feel free to merg!

PR #99, opened 2024-06-04
Lookup for vocab terms without hashmap in RDF reader

### LINKED ISSUE #98: RDF Parser considered harmful (or at least the iteration part is really ugly)
The RDF parser makes multiple passes through both bnode and simple triples many times. And each time
it does so the code is repeated -- something like this.

```rust
for (k, v) in std::mem::take(&mut self.bnode) {
            match v.as_slice() {
                [[_, Term::RDF(VRDF::First), val],
                 [_, Term::RDF(VRDF::Rest), Term::BNode(bnode_id)],
                 // Some sequences have a Type List, some do not
                 ..
                ] => {
                    let some_seq = self.bnode_seq.remove(bnode_id);
                    if let Some(mut seq) = some_seq {
                        seq.push(val.clone());
                        self.bnode_seq.insert(k.clone(), seq);
                        extended = true;
                    } else {
                        self.bnode.insert(k, v);
                    }
                }
                _ => {
                    self.bnode.insert(k, v);
                }
            };
        }
```

I think we can replace this whole thing with a `partition_map` (or a
`map` followed by a partition which would avoid bringing in itertools as a dependency).

Partition map looks like this:

```
partition_map(Iterator<A>, fn -> Either<L, R>) -> (Iterator<L>, Iterator<R>)
```

`L` would be the thing (`Atom`, `Axiom` what ever)
`R` would be `Result<A>`

`L` would then be added to whatever,
`R` would be added back to where ever thing in self it came from in the first place

Would have then:

```rust
let (axiom, triples) = triples.partition_map(
   |t|

   match &t {
      [blah, blah, blah] => {
          // some_ok! would need to return Left or Right(Err), which should be okay
          some_ok!(Left(Axiom(blah, blah, blah))
      }
      None => {
          // Can we take the t here? 
          Ok(Right(&t))
      }
   }
)

self.axioms.extend(axiom);
// We would have to clone the triple here because we only have a reference
self.triples.extend(triples.map(|e| e?.clone()));

// or if we want to capture errors rather than crap out which is something
// we can't do at the moment
let triples, errors = self.triples.partition(is_okay)
self.triples.extend(triples)
self.errors.extend(errors)
```

which is rather neater than what we have at the moment

filippodebortoli:
Leverages the `vocab::Vocab` enum to avoid building and passing around a `Hashmap` when deserializing named nodes in `rdf::reader`.

### DIFF
--- a/src/io/rdf/reader.rs
+++ b/src/io/rdf/reader.rs
@@ -1,5 +1,5 @@
 use rio_api::{
-    model::{BlankNode, NamedNode, Subject},
+    model::{BlankNode, NamedNode, Subject, Triple},
     parser::TriplesParser,
 };
 use Term::*;
@@ -25,10 +25,10 @@ use crate::{
     vocab::RDFS as VRDFS,
 };
 
-use std::collections::HashMap;
 use std::io::BufRead;
 use std::io::Cursor;
 use std::{collections::BTreeSet, convert::TryFrom};
+use std::{collections::HashMap, convert::TryInto};
 
 type RioTerm<'a> = ::rio_api::model::Term<'a>;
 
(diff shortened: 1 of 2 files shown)


phillord:
If I have understood the code correctly, the lookup from `str` to `vocab` enum happens through the vocab module. But, this ultimately backs onto the `find_map` iterator method. If this is correct, it is going to make the lookup linear wrt to the size of the vocab, rather than constant time which the `HashMap` lookup is, although avoiding the cost of building the `HashMap` in the first place.

It's nice that the code is cleaner, but I would worry that this would impact on performance.

filippodebortoli:
> If I have understood the code correctly, the lookup from `str` to `vocab` enum happens through the vocab module. But, this ultimately backs onto the `find_map` iterator method. If this is correct, it is going to make the lookup linear wrt to the size of the vocab, rather than constant time which the `HashMap` lookup is, although avoiding the cost of building the `HashMap` in the first place.
> 
> It's nice that the code is cleaner, but I would worry that this would impact on performance.

That is right.
Right now, `Vocab::FromStr` works as you described.
There is nothing preventing us from changing it to use a `HashMap` lookup.
This can be done e.g. by invoking `Vocab::all()` and collect it into a `HashMap`.
Or by using a more clever way to instantiate the `HashMap` - we could use `std::sync::OnceLock` (successor of `lazy_static!`) to create it similarly to what is done in `enum_meta`, except that the keys of the map are not discriminants, but `str`s.

Overall, I think that cleaning up the code is an important first step, and that changes can be done in `vocab` to ensure that this lookup does not impact too deeply on the parsing runtime. What do you think?

phillord:
This looks good now? It is ready for merge?

filippodebortoli:
> This looks good now? It is ready for merge?

Yes, it is. I will merge now.

PR #102, opened 2024-06-11
RDF reader: Deduplicates code for fetching and processing of triple sequences associated to a BNode.

### LINKED ISSUE #98: RDF Parser considered harmful (or at least the iteration part is really ugly)

filippodebortoli:
This fixes some code duplication issues that were present in the RDF reader.

### DIFF
--- a/src/io/rdf/reader.rs
+++ b/src/io/rdf/reader.rs
@@ -793,25 +793,9 @@ impl<'a, A: ForIRI, AA: ForIndex<A>, O: RDFOntology<A, AA>> OntologyParser<'a, A
                 {
                     ok_some! {
                         {
-                            let facet_seq = self.bnode_seq
-                                .remove(id)?;
-                            let some_facets =
-                                facet_seq.into_iter().map(|id|
-                                                          match facet_map.remove(&id)?.0 {
-                                                              [_, Term::FacetTerm(facet), literal] => Some(
-                                                                  FacetRestriction {
-                                                                      f: facet,
-                                                                      l: self.fetch_literal(&literal)?,
-                                                                  }
-                                                              ),
-                                                              _ => None
-                                                          }
-                                );
-
-                            let facets:Option<Vec<FacetRestriction<_>>> = some_facets.collect();
                             DataRange::DatatypeRestriction(
                                 iri.into(),
-                                facets?
+                                self.fetch_facet_seq(id, &mut facet_map)?
                             )
                         }
                     }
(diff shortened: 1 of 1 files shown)


filippodebortoli:
@phillord this would be ready to merge after review.
I originally only wanted to change io::rdf::reader but there were still some formatting and clippy leftovers to take care of.

filippodebortoli:
@phillord is this PR still relevant after the recent changes to the reader?

phillord:
I am guessing that it doesn't. It would certainly merge badly.

PR #103, opened 2024-06-11
Manchester parser

althonos:
This is the WIP branch for the Manchester syntax, I generated test cases using `bubo` and currently only have the lexer working, it passes all but 6 tests at the moment.

### DIFF
--- a/src/grammars/omn.pest
+++ b/src/grammars/omn.pest
@@ -163,11 +163,12 @@ Annotations             = { "Annotations:" ~ AnnotationAnnotatedList }
 Annotation              = { AnnotationPropertyIRI ~ AnnotationTarget }
 AnnotationTarget        = { Literal | NodeID | IRI }
 
-OntologyDocument  = { SOI ~ PrefixDeclaration* ~ Ontology ~ EOI }
+OntologyDocument  = { SOI ~ Prefixes ~ Ontology ~ EOI }
+Prefixes          = { PrefixDeclaration* }
 PrefixDeclaration = { "Prefix:" ~ PrefixName ~ FullIRI }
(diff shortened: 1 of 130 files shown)


b-gehrke:
Would it be possible to add other extensions to Manchester Syntax (which are implemented in the OWL API) as well? Not all OWL constructs can be expressed in Manchester Syntax when strictly following the OWL standard. The most prominent one being general concept inclusions. We summarized the missing features and extensions in a paper (https://ceur-ws.org/Vol-3515/paper-12.pdf). I'd be happy to help with the implementation!

althonos:
Ah, great that you have a reference document @b-gehrke ! I definitely started to implement some of that because I couldn't get the test cases to pass against the OWL-API output without taking care of the:
```
ObjectProperty:  inverse (r:x)
    EquivalentTo: 
        r:y 
``` 

I'll have a look at the paper and probably rewrite the grammar a bit to account for all possible cases :)

filippodebortoli:
@althonos what is the current status for this PR? Would it be ready for review?

althonos:
@filippodebortoli : I'm not sure, IIRC it was missing a couple of OWL extensions mentioned by Björn, and also I think I encountered a couple of cases making the grammar ambiguous :grimacing: I don't have much time to explore this at the moment unfortunately...

b-gehrke:
@althonos do you have time to look at this PR again? If not I'd be happy to take it on.

althonos:
@b-gehrke : Have a look at this here: https://github.com/fastobo/horned-manchester/ which has actually a more complete version of this PR (including most `FromPair` trait implementation), you can probably reuse most of it

phillord:
I think this is post-hoc now, so will close.

PR #111, opened 2024-06-20
refactor: using &str directly instead of S: Borrow<str>.

filippodebortoli:
* numbers of `clone()`s decreased
* number of explicit `borrow()` calls decreased
* we use `&` a bit more often
* all public methods generating named individuals have a simpler signature


phillord:
I am not convinced by this one. We do simplify some of the API by moving from a generic type to a concrete one, but that increases the complexity in many other areas (lots of "&").

`ForIRI` is now `Deref<str>` so it feels like many of the `borrow` and `clone` could go anyway.

PR #113, opened 2024-06-26
IRI mapped updates for py-horned-owl performance

b-gehrke:
To improve the performance of Py-Horned-OWL, the `BTree(Set|Map)` in `Component`- and `IRIMappedIndex` are changed to `Hash(Set|Map)` and the `IRIMappedOntology` now only holds a `IRIMappedIndex`. The other indexes are now dynamically handled on the Py-Horned-OWL site.

Also, the `SetIndex` is extend by a method to expose its members as smart pointer. When we build additional indexes from a `SetIndex`, this allows us to reuse the memory and not clone the components in the new index.

Currently, I'm having problems with the `IntoIter` implementations for the `IRIMappedIndex`. With the current implementation the tests fail as it just lists all components stored in the index and some components occur multiple times. E.g. `DisjointClasses(:A :B)` occurs once for `:A` and once for `:B`.

### DIFF
--- a/src/ontology/component_mapped.rs
+++ b/src/ontology/component_mapped.rs
@@ -15,7 +15,7 @@ use super::set::SetOntology;
 use crate::model::*;
 use std::{
     cell::RefCell,
-    collections::{BTreeMap, BTreeSet, VecDeque},
+    collections::{HashMap, HashSet, VecDeque},
     iter::FromIterator,
     ops::Deref,
     rc::Rc,
@@ -57,37 +57,37 @@ macro_rules! onimpl {
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
-pub struct ComponentMappedIndex<A, AA> {
-    component: RefCell<BTreeMap<ComponentKind, BTreeSet<AA>>>,
+pub struct ComponentMappedIndex<A: ForIRI, AA: ForIndex<A>> {
+    component: RefCell<HashMap<ComponentKind, HashSet<AA>>>,
     pd: PhantomData<A>,
 }
 
 impl<A: ForIRI, AA: ForIndex<A>> ComponentMappedIndex<A, AA> {
     pub fn new() -> ComponentMappedIndex<A, AA> {
         ComponentMappedIndex {
-            component: RefCell::new(BTreeMap::new()),
+            component: RefCell::new(HashMap::new()),
             pd: Default::default(),
         }
     }
 
     /// Fetch the component hashmap as a raw pointer.
     ///
-    /// This method also ensures that the BTreeSet for `cmk` is
+    /// This method also ensures that the HashSet for `cmk` is
     /// instantiated, which means that it effects equality of the
     /// ontology. It should only be used where the intention is to
     /// update the ontology.
-    fn component_as_ptr(&self, cmk: ComponentKind) -> *mut BTreeMap<ComponentKind, BTreeSet<AA>> {
+    fn component_as_ptr(&self, cmk: ComponentKind) -> *mut HashMap<ComponentKind, HashSet<AA>> {
         self.component.borrow_mut().entry(cmk).or_default();
         self.component.as_ptr()
     }
 
     /// Fetch the components for the given kind.
-    fn set_for_kind(&self, cmk: ComponentKind) -> Option<&BTreeSet<AA>> {
+    fn set_for_kind(&self, cmk: ComponentKind) -> Option<&HashSet<AA>> {
         unsafe { (*self.component.as_ptr()).get(&cmk) }
     }
 
     /// Fetch the components for given kind as a mutable ref.
-    fn mut_set_for_kind(&mut self, cmk: ComponentKind) -> &mut BTreeSet<AA> {
+    fn mut_set_for_kind(&mut self, cmk: ComponentKind) -> &mut HashSet<AA> {
         unsafe { (*self.component_as_ptr(cmk)).get_mut(&cmk).unwrap() }
     }
 
@@ -203,7 +203,7 @@ onimpl! {AnnotationPropertyDomain, annotation_property_domain}
 onimpl! {AnnotationPropertyRange, annotation_property_range}
 onimpl! {Rule, rule}
 
-impl<A, AA> Default for ComponentMappedIndex<A, AA> {
+impl<A: ForIRI, AA: ForIndex<A>> Default for ComponentMappedIndex<A, AA> {
     fn default() -> Self {
         Self {
             component: Default::default(),
@@ -217,12 +217,12 @@ impl<A: ForIRI, AA: ForIndex<A>> IntoIterator for ComponentMappedIndex<A, AA> {
     type Item = AnnotatedComponent<A>;
     type IntoIter = std::vec::IntoIter<AnnotatedComponent<A>>;
     fn into_iter(self) -> Self::IntoIter {
-        let btreemap = self.component.into_inner();
+        let hashmap = self.component.into_inner();
 
         // The collect switches the type which shows up in the API. Blegh.
-        let v: Vec<AnnotatedComponent<A>> = btreemap
+        let v: Vec<AnnotatedComponent<A>> = hashmap
             .into_values()
-            .flat_map(BTreeSet::into_iter)
+            .flat_map(HashSet::into_iter)
             .map(|fi| fi.unwrap())
             .collect();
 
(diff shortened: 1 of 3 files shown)


phillord:
Could you rebase against devel?

I've added criterion benching there and am using this to check changes for performance. I've updated Build to use HashSet which seems to have mostly (but not entirely) positive impact on performance. And, I've updated `ComponentMapped` and am benchmarking this at the moment. 

Losing iteration order from `ComponentMapped` definately makes some things harder (i.e. I have to sort the tests!), so it is something that I don't want to do unless it definately helps. Same will be true for `IRIMapped`.

phillord:
Well I have checked switching from BTreeSet to HashSet for ComponentMappedOntology.  Despite my expectations, it seems to slow things considerably.

Hopefully you can see the graph below. Given the complexity this adds with lack of sort order, it looks like a poor optimisation.

https://bencher.dev/perf/horned-owl?lower_value=false&upper_value=false&lower_boundary=false&upper_boundary=false&x_axis=date_time&branches=9f82f2c6-9086-468e-8ca9-1e30c2c8e362&testbeds=4328351a-d4a0-4872-8bba-fa5052f9db27&benchmarks=f114f934-8042-4d12-b41d-e953a5187c47%2Cf5305925-eb01-4006-baa5-0a32aacca3b0%2C53d7104a-3c10-4707-b067-9f0595bb1ab6%2C9ee24530-9892-49ae-8b3c-c047ec9b0fc0%2C93c39a32-cd5e-4b50-b971-11d24fb8286f%2C035c499c-1a99-432e-8509-dd91134019c7%2C12da761f-d6f9-4516-9625-89c8b44609ca%2C86aec33a-3a13-4718-ad53-31208f43f025&measures=02bafc13-19a5-46c5-8bf1-8d346880443d&start_time=1712320149655&end_time=1719577749655&tab=plots&plot=86d8bef8-0b43-4c58-915e-065b230afa88&plots_search=86d8bef8-0b43-4c58-915e-065b230afa88&key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&plots_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=1&plots_page=1

phillord:
I am wondering whether part of the problem here is that the read methods return concrete types. If `owx::read` were generic over the return type, we could have it return a `PyIndexOntology` directly. The RDF reader would have to return a type with at least the ability to return `DeclarationMapped` indexes. And during tests, we could return ontologies with `ComponentMappedIndex` directly rather than converting.

I am not sure if the type system is rich enough to allow this, but it might be worth experimenting.

b-gehrke:
> Well I have checked switching from BTreeSet to HashSet for ComponentMappedOntology. Despite my expectations, it seems to slow things considerably.

Interesting results. I would've expected HashSet to be more performant than BTreeSet as well. But I'll change it back to BTreeSet then.

phillord:
I guess HashSet is re-allocating a lot as it grows; just bunging in a bigger default might solve this. You can see the previous data points are the use of HashSet for Build -- again, not a simple "everything gets faster" result.

Before removing HashSet, let me benchmark your code.

b-gehrke:
> I am wondering whether part of the problem here is that the read methods return concrete types. If owx::read were generic over the return type, we could have it return a PyIndexOntology directly.

I'm not sure if this is really a problem. When we have access to the underlying index(es) we can just move them to the new ontology in a `From` implementation.

But I'm wondering why `SetOntology` is only generic in the IRI part and not the AnnotatedComponent part.

phillord:
HashSet considered harmful again!


https://bencher.dev/perf/horned-owl?lower_value=false&upper_value=false&lower_boundary=false&upper_boundary=false&x_axis=date_time&branches=c366e694-a93b-448e-b85c-bfd99d083de3&testbeds=4328351a-d4a0-4872-8bba-fa5052f9db27&benchmarks=0989f2b4-c019-427d-a323-d4c688fc2193%2Cd44dcd4e-4c87-4b73-972b-b742c86246ed%2C112415c0-2509-44df-b34f-464ebf6b517a%2C7dd911d5-14d2-4e60-ade4-d0747605c422%2C4d1c63a0-d388-4d85-83df-331c9538d34c&measures=02bafc13-19a5-46c5-8bf1-8d346880443d&start_time=1717169901650&end_time=1719589101650&tab=plots&plots_search=a6db6fa3-69b6-4ec3-84ac-6b1f089469aa&key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&plots_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=1&plots_page=1

phillord:
In answer to the question about `SetOntology`, it's backed by `SetIndex` which is generic for `A` and `AA`. If you want that genericity then `OneIndexedOntology<SetIndex<A, AA>` gives it to you.  


Before I created the indexing method I only had `SetOntology` and that was directly backed by a `HashSet`. I wanted to retain something that was a fast as this when I created `SetIndex` -- that wasnt going to happen if `SetIndex` always uses an `Rc`, but fortunately the `AA` generic means I do not need to use this.  And, by providing an new type wrapper means that you don't need to specify the `AA` generic, so it's one less type bound to type. Again, something I wanted to keep when adding Indexes -- the simple case should not get harder. So now I have all of this, but at the cost that the `SetOntology` feels a bit asymmetric -- hence your question. 


If you want to convert from another `Index` instance to a set backed ontology without unwinding all the `Rc`s, the use `OneIndexedOntology<A, AA>`

phillord:
Something like this:

https://github.com/phillord/horned-owl/tree/feature/generic-return-owx-reader


Now, we can return any type from the OWXReader.  It should be able to return a py horned ontology directly.

The RDF parser is a bit harder and I think that the API needs cleaning up a bit elsewhere. But would this fulfil your need?

b-gehrke:
Yes, that works very well! With that, Py-Horned-OWL is ~30% slower than horned-owl. While still better than before, I cannot believe that these 30% are Arc and PyO3 overhead. But with your changes we do literally the same as Horned-OWL. I'll do some more benchmarking...

Would it be possible for the ofn reader to be generic too?

The RDF reader is not a problem for me at the moment. Its set index already stores Arcs, so we simply move the index from the RDFOntology to the PyIndexedOntology.

b-gehrke:
Ah, I forgot to enable all optimizations. We actually have 10% performance loss now. And using a profiler, I can see that the actual parse time is roughly the same. But PyO3s finalizing / deallocating memory takes double the time.. I'm not sure why.

phillord:
I haved checked `Arc`/`Rc` (run cargo bench and look at the "tree" and "food" benchmarks. Bottom line `Arc` is a bit slower but single digit percentile differences in worst case. In most cases, immeasurable differences (to the point that I wonder if adding all the generics was worth it when I could just have done a search and replace on `Rc`!).

The RDF reader is generic for `read_with_build` over `A` and `AA` but not fully generic. Perhaps that is enough. The `read` method is hard coded to `RcStr`. In the tests, the first thing I do is dump all of these to `SetOntology`, but that makes the indexing simpler. So you are probably correct, more generics is not needed.

OFN, don't know, don't see why not, but will look.

In terms of the overhead for PyO3, have you tried launching it without parsing an ontology, or parsing an empty ontology. That would give you a test for the best you can get.

phillord:
RDF now also generic on that branch! Rather harder than OWX but now I know how to do it for anything.

phillord:
And OFN. @althonos can you take a look? I had to fiddle with the internal API a bit. But it makes your tests simpler because you can dump out a ComponentMappedOntology directly which you need for the writer.

I am benchmarking this at the moment, but it should not have an impact. Really we need to update the writers also to work in terms of `AsRef<Index>` calls.

phillord:
After we have merged #115, do we still need this or part of this?

If so, we will need to rebase what remains and do a new PR. Either way, can you close this one if you are happy.

b-gehrke:
I certainly don't need all of it - I'll have look. But I still think the `IRIMappedOntology` should only contain a `IRIMappedIndex`. After all, it is what I'd expect from the name.

But we now store the indexes directly in Py-Horned-OWLs PyIndexedOntology. So, strictly speaking, we wouldn't need the `IRIMappedOntology` anymore and it could be removed.

Do you think the `IntoIterator` implementation is sensible? I thought about it returning an `Iter<Item=(IRI, BTreeSet<AnnotatedComponent<A>>)>` instead of `Iter<Item=AnnotatedComponent<A>>`. It would make it clearer to the user why some components are stored multiple times but it would break with the other indexes which return an iterator over annotated components. We could also remove the duplicates, but this would add another  overhead and would not reflect what is actually stored in the index. What are your thoughts?

phillord:
I agree with your point `IRIIndexOntology`. I need to think about how many different `Ontology` we need. Having lots of newtypes isn't the greatest idea and both `SetOntology` and `ComponentMappedOntology` are new types around `OneIndexed` and as your earlier question shows `SetOntology` is confusingly so. I introduced it for historical, backward compact purposes. But newtypes are a pain and need lots of trait implementations.

It's something to put on the list for Horned-OWL 2.0. A consistent approach to when to have newtypes and when not.

b-gehrke:
The only thing that would be good to have is an `iter` method on the SetIndex. I'll create another PR for that.

PR #121, opened 2024-11-14
Remove component for all IRIs from IRIMappedIndex

b-gehrke:
Previously, `index_remove` and `index_take` removed only the components for the first IRI in a component from an `IRIMappedIndex` index.

### DIFF
--- a/src/ontology/iri_mapped.rs
+++ b/src/ontology/iri_mapped.rs
@@ -199,25 +199,17 @@ impl<A: ForIRI, AA: ForIndex<A>> OntologyIndex<A, AA> for IRIMappedIndex<A, AA> 
     }
 
     fn index_take(&mut self, cmp: &AnnotatedComponent<A>) -> Option<AnnotatedComponent<A>> {
-        let iris = self.aa_to_iris(cmp);
-        if !iris.is_empty() {
-            let iri = iris.iter().next();
-            if let Some(iri) = iri {
-                self.mut_set_for_iri(iri).take(cmp).map(|aax| aax.unwrap())
-            } else {
-                None
-            }
-        } else {
-            None
-        }
+        self.aa_to_iris(cmp).iter().fold(None, |val, iri| {
+            self.mut_set_for_iri(iri)
+                .take(cmp)
+                .map_or(val, |c| Some(c.unwrap()))
+        })
     }
 
     fn index_remove(&mut self, cmp: &AnnotatedComponent<A>) -> bool {
-        if let Some(iri) = self.aa_to_iris(cmp).iter().next() {
-            self.mut_set_for_iri(&iri.clone()).remove(cmp)
-        } else {
-            false
-        }
+        self.aa_to_iris(cmp).iter().fold(false, |val, iri| {
+            self.mut_set_for_iri(iri).remove(cmp) || val
+        })
     }
 }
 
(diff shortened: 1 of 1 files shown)


phillord:
Looks find to me. Can you squash the formatting changes ?

b-gehrke:
Sure, done.

phillord:
Feel free to merge to devel

src/ontology/iri_mapped.rs
```diff
@@ -199,25 +199,18 @@ impl<A: ForIRI, AA: ForIndex<A>> OntologyIndex<A, AA> for IRIMappedIndex<A, AA>
...
-            if let Some(iri) = iri {
-                self.mut_set_for_iri(iri).take(cmp).map(|aax| aax.unwrap())
-            } else {
-                None
-            }
-        } else {
-            None
-        }
+        self.aa_to_iris(cmp).iter().fold(None, |val, iri| {
+            self.mut_set_for_iri(&iri.clone())
```
filippodebortoli: `&iri.clone()` could be safely replaced with `iri` to avoid unnecessary memory allocation.

src/ontology/iri_mapped.rs
```diff
@@ -199,25 +199,18 @@ impl<A: ForIRI, AA: ForIndex<A>> OntologyIndex<A, AA> for IRIMappedIndex<A, AA>
...
     }
 
     fn index_remove(&mut self, cmp: &AnnotatedComponent<A>) -> bool {
-        if let Some(iri) = self.aa_to_iris(cmp).iter().next() {
-            self.mut_set_for_iri(&iri.clone()).remove(cmp)
-        } else {
-            false
-        }
+        self.aa_to_iris(cmp).iter().fold(false, |val, iri| {
+            self.mut_set_for_iri(&iri.clone()).remove(cmp) || val
```
filippodebortoli: `&iri.clone()` could be safely replaced with `iri` to avoid unnecessary memory allocation.

src/ontology/iri_mapped.rs
```diff
@@ -199,25 +199,18 @@ impl<A: ForIRI, AA: ForIndex<A>> OntologyIndex<A, AA> for IRIMappedIndex<A, AA>
...
-            } else {
-                None
-            }
-        } else {
-            None
-        }
+        self.aa_to_iris(cmp).iter().fold(None, |val, iri| {
+            self.mut_set_for_iri(&iri.clone())
+                .take(cmp)
+                .map(|c| c.unwrap())
```
filippodebortoli: Why not using `map_or()` instead of `.map().or()`?

filippodebortoli (approved):
The changes in control flow are OK, but there is some unnecessary cloning happening. I marked the lines where this could be avoided.

src/ontology/iri_mapped.rs
b-gehrke: For me it feels more natural than to wrap the return value in `Some`. Is it cleaner to write `.map_or(val, |c| Some(c.unwrap()))` instead?

filippodebortoli: Probably not, if you then have to wrap again. I will mark this line as resolved.

src/ontology/iri_mapped.rs
b-gehrke: Done

src/ontology/iri_mapped.rs
b-gehrke: Done

filippodebortoli (approved):

PR #122, opened 2024-11-21
Documentation for `ontology::iri_mapped`

filippodebortoli:
Adding missing documentation to this module.

- [ ] basic method documentation: functionality, panics, assumptions.
- [ ] usage examples for public methods.

### DIFF
--- a/src/ontology/iri_mapped.rs
+++ b/src/ontology/iri_mapped.rs
@@ -24,6 +24,31 @@ use super::set::SetIndex;
 
 use std::collections::HashSet;
 
+/// Implements an ontology index that stores for each IRI a set of annotated components in which this IRI occurs.
+///
+/// # Examples
+///
+/// The following snippet shows the effect of adding an axiom involving two IRIs to a new index.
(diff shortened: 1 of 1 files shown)


phillord:
You have the "RefCell" commit on this PR, but that has already gone to devel, so think you are going to get a conflict when you try and merge.

filippodebortoli:
> You have the "RefCell" commit on this PR, but that has already gone to devel, so think you are going to get a conflict when you try and merge.

I was trying to rebase the changes I applied on top of `devel` and somewhat I reapplied that commit. I now reset and the commit won't be re-applied - I may have to do some merge resolution, but I am fine with it.

*EDIT* I managed to correctly rebase on `devel`, now the merge should happen automatically and without reapplying any commit. @phillord is there any other information that I should add as documentation to this module?

phillord:
I think that this looks good to me -- I would probably squash the two commits into one as they are basically the same thing, but it is not a big issue either way. Do feel free to push to devel.

Issue #126, opened 2025-02-18
How to share an ArcIRIMappedOntology between threads?


KonradHoeffner:
Our application loads an ontology on startup and then serves a view of it to multiple clients using multiple threads.
This compiles without problem when using a `SetOntology<Arc<str>>>` but now I need an index as well and it fails to build.
I'm not an expert with threads in Rust but until now I just always used Arc and it worked.
Is this because the ArcIRIMappedOntology uses RefCell internally, which is not thread safe?
And is this in turn because it is mutable?
As I don't need mutability, is there any way of making it immutable and then sharing it?
Sorry if this is a basic question.

```rust
static COMPILES: OnceLock<SetOntology<Arc<str>>> = OnceLock::new();
static FAILS: OnceLock<ArcIRIMappedOntology> = OnceLock::new();   
```

```
error[E0277]: `RefCell<BTreeMap<ComponentKind, BTreeSet<Arc<AnnotatedComponent<Arc<str>>>>>>` cannot be shared between threads safely
   --> src/rdf.rs:86:15
    |
86  | static FAILS: OnceLock<ArcIRIMappedOntology> = OnceLock::new();
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `RefCell<BTreeMap<ComponentKind, BTreeSet<Arc<AnnotatedComponent<Arc<str>>>>>>` cannot be shared between threads safely
    |
    = help: within `IRIMappedOntology<Arc<str>, Arc<AnnotatedComponent<Arc<str>>>>`, the trait `Sync` is not implemented for `RefCell<BTreeMap<ComponentKind, BTreeSet<Arc<AnnotatedComponent<Arc<str>>>>>>`
    = note: if you want to do aliasing and mutation between multiple threads, use `std::sync::RwLock` instead
note: required because it appears within the type `ComponentMappedIndex<Arc<str>, Arc<AnnotatedComponent<Arc<str>>>>`
   --> /home/konrad/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/horned-owl-1.0.0/src/ontology/component_mapped.rs:59:12
    |
59  | pub struct ComponentMappedIndex<A, AA> {
    |    
```

phillord:
Yes, both `component_mapped` and `iri_mapped` use `RefCell` which isn't sync, so it can't be shared. 

[...]

PR #127, opened 2025-02-20
ontology: removes RefCells from indexes.

filippodebortoli:
Addresses #126.

It is now possible to wrap `ArcIRIMappedOntology` using `OnceLock`, `Mutex` etc.
Additionally, this change removes some unsafe code that was needed to access the `BTreeMap` that was wrapped in a `RefCell`. 

I added `test_once_lock_ontology()` to mimic the behavior showed by @KonradHoeffner in #126, which now passes.

### DIFF
--- a/src/ontology/component_mapped.rs
+++ b/src/ontology/component_mapped.rs
@@ -14,7 +14,6 @@ use super::indexed::ForIndex;
 use super::set::SetOntology;
 use crate::model::*;
 use std::{
-    cell::RefCell,
     collections::{BTreeMap, BTreeSet, VecDeque},
     iter::FromIterator,
     ops::Deref,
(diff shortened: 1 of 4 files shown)


phillord:
Looks good to me. Both simpler and more functional!

Guess I had not discovered the `Entry` api when I wrote this. That `unsafe` code always irritated me, but I couldn't see a way around. I will be happy to see it gone.

filippodebortoli:
@phillord merged. Should we release a minor version to account for this fix?

KonradHoeffner:
Wow, thank you for the super fast accommodation of our requirements, this is much appreciated!
A minor version release would be perfect on our end.
I hope this doesn't break existing code for other users.

Issue #124, opened 2025-02-18
RDF/XML fails to parse when using OWL punning


KonradHoeffner:
When loading an RDF/XML ontology with 49k lines with the newest release 1.0.0 of horned-owl, I get an "not yet implemented" error.
However the error message does not indicate what exactly the problem is.
I think it would help the users if the error messages would be more descriptive.
Fictitious example "Error parsing line 12345 in RDF/XML ontology: named entities with URIs longer than 1000 characters are not supported".

Or am I loading the ontology wrong? Another file can be loaded successfully.

```rust
fn load_ontology() -> Result<SetOntology<ArcStr>, HornedError> {
    let b = Build::<ArcStr>::new();
    let mut br = kb_reader("data/snik.rdf").unwrap(); // returns a buffered reader
    Ok(horned_owl::io::ParserOutput::<Arc<str>, ArcAnnotatedComponent>::rdf(horned_owl::io::rdf::reader::read_with_build(
        &mut br,
        &b,
        ParserConfiguration::default(),
    ).unwrap())
    .decompose()                                                              
    .0)
}
```

```
thread 'actix-rt|system:0|arbiter:1' panicked at /home/konrad/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/horned-owl-1.0.0/src/io/rdf/reader.rs:1672:30:
not yet implemented
```

2a9da0aa9747f4042af60e9d1a2a0caa6b549099/src/io/rdf/reader.rs#L1672C1

```rust
        [Term::Iri(sub), Term::Iri(pred), Term::Iri(obj)] => ok_some! {
                    match (self.find_declaration_kind(sub, ic)?,
                           self.find_declaration_kind(pred, ic)?,
                           self.find_declaration_kind(obj, ic)?) {
                        (NamedOWLEntityKind::NamedIndividual,
                         NamedOWLEntityKind::ObjectProperty,
                         NamedOWLEntityKind::NamedIndividual) => {
                            ObjectPropertyAssertion {
                                ope: ObjectProperty(pred.clone()).into(),
                                from: sub.into(),
                                to: obj.into()
                            }.into()
                        }
                        _ => todo!()
                    }
                },
                _ => Ok(None),
            };
```

The file causing the error : [snik.rdf compressed as snik.zip](https://github.com/user-attachments/files/18842642/snik.zip)

phillord:
[...]

PR #130, opened 2025-02-21
Support Punning, resolve snik failure

phillord:
This branch will fully support punning and resolve the problems with this causes with parsing snik.

https://github.com/phillord/horned-owl/issues/124


phillord:
I believe that the unit test I have added is now showing the same issue.

phillord:
My proposed solution is this, which I think will fix the problem. I think it makes sense to refactor the "find_declaration_kind" methods to make them utility function of declaration_mapped at the same time.


 - [ ] Add method declaration_kind_pun_as(iri, NamedEntityKind) -> NamedEntityKind
 - [ ] Add utility "closure_declaration_kind" to Declarationmappedindex
 - [ ] Add utility "closure_declaration_kind_pun_as"
 - [ ] Refactor RDF reader to use these

filippodebortoli:
Naive question: what would go wrong if we treated punning by changing `DeclarationMappedIndex` so that it wraps `HashMap<IRI<A>, Vec<NamedEntityKind>>` instead of `HashMap<IRI<A>, NamedEntityKind>`? Having a vector would allow us to store all the possible types that an entity with a certain IRI assumes in an ontology.

Then, if we want to check if `iri` is declared as `kind` and extract the corresponding named entity, we could introduce e.g. `get_object_property(&self, &iri) -> Option<ObjectProperty>`.

Then,

```rust
    fn fetch_ope(&mut self, t: &Term<A>, ic: &[&O]) -> Option<ObjectPropertyExpression<A>> {
        match self.find_property_kind(t, ic)? {
            PropertyExpression::ObjectPropertyExpression(ope) => Some(ope),
            _ => None,
        }
    }
```

would become

```rust
    fn fetch_ope(&mut self, t: &Term<A>, ic: &[&O]) -> Option<ObjectPropertyExpression<A>> {
        self.get_object_property(t, ic).map(|op| op.into())
    }
```

Is this along the line of what you suggested? Or is it different and perhaps not working as intended?

phillord:
@filippodebortoli  I think it would hit more code, that's all. Punning is only valid for classes and individuals. The rest of the time, we want to make decisions on the basis of whether something is, for example, an object or data property. If `kind` can return several answers, we would have to deal with that hassle all the time.

Of course, this means that we cannot use declaration mapped index to do things like all IRIs that have been declared (illegally) to be two different types. If we wanted to do this kind of validation, though, we could write a new index.

phillord:
Well, the commits now suggest that we support punning. But we still fail to parse snik. 


The problem is that puns are not all declared. For example, this definition does declare both a class and an individual.

````
  <ciox:Finanzwesensystem rdf:about="http://www.snik.eu/ontology/ciox/SAPFI">
    <meta:supports rdf:resource="http://www.snik.eu/ontology/ciox/Rechnungspruefung"/>
    <rdf:type rdf:resource="http://www.snik.eu/ontology/meta/EntityType"/>
    <rdf:type rdf:resource="http://www.w3.org/2002/07/owl#Class"/>
    <rdf:type rdf:resource="http://www.w3.org/2002/07/owl#NamedIndividual"/>
    <rdfs:label xml:lang="de">SAP FI</rdfs:label>
    <rdfs:label xml:lang="en">SAPFI</rdfs:label>
    <rdfs:subClassOf rdf:resource="http://www.snik.eu/ontology/ciox/SAP"/>
  </ciox:Finanzwesensystem>
````

And the declarationmappedindex correctly identifies it as such. However, the crash is caused by this triple:

Error on PosTriple([Iri(IRI("http://www.snik.eu/ontology/ciox/Abteilungsleiter")), Iri(IRI("http://www.snik.eu/ontology/meta/isInvolvedIn")), Iri(IRI("http://www.snik.eu/ontology/ciox/CABDurchfuehrung"))], 1108149), (Class, ObjectProperty, Class)


````        
    <meta:Role rdf:about="http://www.snik.eu/ontology/ciox/Abteilungsleiter">
    <meta:isInvolvedIn rdf:resource="http://www.snik.eu/ontology/ciox/CABDurchfuehrung"/>
    <meta:isInvolvedIn rdf:resource="http://www.snik.eu/ontology/ciox/DurchfuehrungJourFixe"/>
    <meta:isInvolvedIn rdf:resource="http://www.snik.eu/ontology/ciox/Projektdurchfuehrung"/>
    <meta:isInvolvedIn rdf:resource="http://www.snik.eu/ontology/ciox/Projektplanung"/>
    <meta:isInvolvedIn rdf:resource="http://www.snik.eu/ontology/ciox/Rechnungspruefung"/>
    <rdf:type rdf:resource="http://www.w3.org/2002/07/owl#Class"/>
    <rdfs:label xml:lang="de">Abteilungsleiter</rdfs:label>
    <rdfs:label xml:lang="en">Department Manager</rdfs:label>
    <rdfs:subClassOf rdf:resource="http://www.snik.eu/ontology/ciox/InformationManagementStaff"/>
    <skos:closeMatch rdf:resource="http://dbpedia.org/resource/Abteilungsleiter"/>
  </meta:Role>
 ````

And, here there is no definition of this as a individual only a class.  So, Horned-OWL does not know what to do and 
crashes.

So, I need to think about what is the right thing to do here. Following directly on from the specification, I think, that Horned's behaviour is reasonable, just not very useful.

Thoughts welcome.

**Comment by KonradHoeffner** [comment]
> Of course, this means that we cannot use declaration mapped index to do things like all IRIs that have been declared (illegally) to be two different types. If we wanted to do this kind of validation, though, we could write a new index.

Could this be an "Err" Result of the declaration index constructor?
Alternatively, the first encountered type (or one chosen from a priority list) could be used and warnings could be logged and/or returned as an additional returned object.
This would have the advantage that "faulty" ontologies that the user does not own could still be used and if they are large those "faulty" parts may not even be relevant to the user queries.

The latter approach is used by Protégé, which when loading SNIK outputs a red message to the console but still loads the file:
```
Illegal redeclarations of entities: reuse of entity http://www.snik.eu/ontology/it4it/ID in punning not allowed [Declaration(DataProperty(<http://www.snik.eu/ontology/it4it/ID>)), Declaration(AnnotationProperty(<http://www.snik.eu/ontology/it4it/ID>))] 
```

I think for the user this is the best approach, because for example in SNIK, we have different subontologies where most are verified and validated, but some are also unfinished and rarely used.
As SNIK is developed in RDF Turtle files and validated using SHACL, and not primarily used with OWL tools like Protégé, some OWL errors may be in there but I'm not sure how representative that is for other knowledge bases and ontologies.

**Comment by phillord** [comment]
The difficult question is what we mean by error.  The OWL api reports a large number of OWL2DL profile declarations (either illegal puns or undeclared entities, both classes and annotation properties). But this does not mean it is not legal OWL full.

I will push a version soon that can parse snik.rdf. But it has to make some assumptions to do this and currently leaves over ~800 unparsed triples which I don't understand at the moment.

I think I will need something similar to the OWL API which is a strict mode (which will error) and an lax mode (which will make assumptions). The bones of this are already there so it should not be too bad.

At the moment, all of the test cases for Horned-OWL are generated by the OWL API; so it is biased toward OWL that is generated by the OWL API, rather than arbitrary tools. I will have to think about this also.

**Comment by phillord** [comment]
This commit now means that snik is parsed without crashing. 

However, there are still a substantial number of unparsed triples.

````
Ontology has:
	Logical Components: 41782
	Annotation Components: 0
	Meta Components: 2

Detailed
	Ontology ID                              | 1    
	Doc IRI                                  | 1    
	Ontology Annotation                      | 816  
	Declare Class                            | 2216 
	Declare Object Property                  | 66   
	Declare Annotation Property              | 20   
	Declare Data Property                    | 270  
	Declare Named Individual                 | 36   
	Sub-Class Of                             | 3384 
	Equivalent Classes                       | 5    
	Disjoint Classes                         | 6    
	Sub Object Property Of                   | 34   
	Object Property Domain                   | 57   
	Object Property Range                    | 50   
	Inverse Functional Object Property       | 1    
	Irreflexive Object Property              | 1    
	Symmetric Object Property                | 2    
	Data Property Domain                     | 23   
	Data Property Range                      | 26   
	Class Assertion                          | 4734 
	Object Property Assertion                | 1659 
	Annotation Assertion                     | 28371
	Sub Annotation Property Of               | 5    


Incompleted Parsed
	Simple Triples: 838
	Bnode: 28
	Sequences: 7
	Class Expressions: 0
	Object Property Expressions: 0
	Data Range: 0
	Atom: 0
	Annotations: 0

````

If I try the same thing with snik that has been sanitised through the OWL API I get a much cleaner parse.

````
Ontology has:
	Logical Components: 48109
	Annotation Components: 42
	Meta Components: 2

Detailed
	Ontology ID                              | 1    
	Doc IRI                                  | 1    
	Ontology Annotation                      | 4750 
	Declare Class                            | 3900 
	Declare Object Property                  | 69   
	Declare Annotation Property              | 69   
	Declare Data Property                    | 270  
	Declare Named Individual                 | 4818 
	Declare Datatype                         | 2    
	Sub-Class Of                             | 3388 
	Equivalent Classes                       | 5    
	Disjoint Classes                         | 3    
	Sub Object Property Of                   | 35   
	Object Property Domain                   | 50   
	Object Property Range                    | 50   
	Inverse Functional Object Property       | 1    
	Irreflexive Object Property              | 1    
	Symmetric Object Property                | 2    
	Data Property Domain                     | 6    
	Data Property Range                      | 9    
	Class Assertion                          | 4734 
	Object Property Assertion                | 7695 
	Data Property Assertion                  | 970  
	Annotation Assertion                     | 17277
	Sub Annotation Property Of               | 5    


Incompleted Parsed
	Simple Triples: 52
	Bnode: 6
	Sequences: 0
	Class Expressions: 0
	Object Property Expressions: 0
	Data Range: 0
	Atom: 0
	Annotations: 0
````

The numbers here are still quite different from the values produced by robot and the OWL API, but that is perhaps to be expected.

I will next look at the unparsed triples and see what I can do there. I think we do need to accept that both horned and the OWL API think snik require a reasonable degree of guesswork to happen.

@KonradHoeffner Do you have metrics for snik? How many classes do you think it has?

**Comment by KonradHoeffner** [comment]
@phillord: Thank you for all the work! Yes, we have a lot of metrics for SNIK, as we built [a whole dashboard for them](https://snikproject.github.io/blog/2017/04/12/dashboard/), key stats are:

| | |
|-------|--------|
|classes|6353|
|properties|336|
|triples|41732|

**Comment by phillord** [comment]
Hmmm. Well, 6000 classes is a value that neither the OWL API nor Horned gets to when looking at snik.rdf. Is it possible to see the code that you have used to calculate this, so I can see how you are coming to this value?

I fear that I am coming to the conclusion that quite a few of the triples in snik are not valid against the OWL2 specification. So there is always going to be a degree of guess work.

**Comment by KonradHoeffner** [comment]
I think the problem is that while we call SNIK an ontology, in fact it is more of a hybrid between an ontology and a knowledge base.
While our goal is to be valid OWL 2, our whole pipeline is RDF-based with validation based on SHALC, SPARQL queries and Redland Raptor (rapper).
We just always assumed that if we produce valid RDF/XML or Turtle and our SHACL shapes validate and it loads in Protégé without displaying an error or a warning that we have valid OWL 2 as well but it seems we have to revise that assumption and add an OWL 2 validator to our pipeline as well.
However some of the subontologies that are unfinished still have validation errors because the people working on them didn't have time finish them when we finished the project, I will try to make them valid OWL 2 at least.
So we will try to fix as many of those validation errors as we can and then try it again.
As for the number of classes, we count them as follows on <https://www.snik.eu/sparql>:

```sparql
SELECT (COUNT(?class) AS ?classes) FROM <http://www.snik.eu/ontology>
{
 {?class a owl:Class.} UNION {?class a [rdfs:subClassOf meta:Top]}.
}
```

So for example in SNIK we would have:

```ttl
meta:Top a owl:Class.
meta:Role rdfs:subClassOf meta:Top.

bb:HisStakeholder a meta:Role.

bb:HospitalStaff a meta:Role; rdfs:subClassOf bb:HisStakeholder.

bb:ChiefExecutiveOfficer
    a meta:Role ;
    rdfs:subClassOf bb:HospitalStaff .
```

The reason we modelled it this way because our focus was too use as few extra triples as possible and allow applications to use SPARQL queries that are as simple as possible (no property path queries, just rdf:type to figure out what main type something is to determine the right shape to display) but it seems I went overboard and destroyed the OWL validity in the process. Sorry if this created extra work for you, I will investigate how we can fix this in the easiest way, it seems as if the most important thing is for us to add the missing rdfs:subClassOf statements.

phillord:
It's an interesting question. I guess you would need to distinguish between OWL2 DL and OWL2 full. A number of the statements made in snik.rdf could, I guess, be interpreted into OWL full. And, perhaps, horned should parse them, although my parser is, as written, perhaps not ideally supported to do this.

I'd be willing to consider fixing this but then, the question is what meaning you are giving to the statements that are currently not parsing. I don't want to go to the effort of supporting something that was not intended. 

Horned will work as a good OWL validator; happy to help you integrate it into your pipelines if you want.

phillord:
Although snik does not fully parse, it does parse without error and this now supports punning. So I think I will merge this and leave full parsing of snik for another PR.

phillord:
Closed in favour of https://github.com/phillord/horned-owl/pull/160

Issue #129, opened 2025-02-20
Fix todo! statements


phillord:
There are a number of todo! statements in horned which should be removed or replaced with panics or unreachables where they are.

Related: https://github.com/phillord/horned-owl/issues/124

cthoyt:
Hi, I'm just running into this now while using py-horned-owl, and unfortunately, since there isn't any information in most TODOs, it gives incomprehensible error messages. I would love to see minimally explanations of the objects that are panic'd instead of blank todos

phillord-ncl:
If you know which todo! is giving you a problem, then please do let me know and I can prioritize.

cthoyt:
cheers @phillord i will send that later today. After I see how you address one, hopefully I can contribute for other ones. I also have to set up py-horned-owl to compile with a fork of horned-owl, which I will figure out as a rust newbie :)

phillord:
Putting something vaguely useful in the todo! or panic! output would be a very good intermediate step. In the ideal world, they need rewriting with errors, but that will take a bit more fiddling,

cthoyt:
This is the one that broke for me https://github.com/phillord/horned-owl/blob/0477f81ef2047889c946e0cb1392ec163728a259/src/io/rdf/reader.rs#L1785 when loading https://github.com/biopragmatics/obo-db-ingest/raw/refs/heads/main/export/ec/ec.owl. 

In general, I will try to test this on all ontologies in https://github.com/biopragmatics/obo-db-ingest

phillord:
https://github.com/phillord/horned-owl/pull/160

This is a trivial fix that updates the todo! statement to show something useful. 

Why this doesn't work at all, I will need to investigate a bit more!

phillord:
And now updated to actually do proper error handling. 

The error comes because Horned is interpreting this IRI `https://www.enzyme-database.org/query.php?ec=1.1.1.n11` as a class, hence the triple at point is "Class->ObjectProperty->NamedIndividual".  I am guessing that the intention is that the IRI is interpreted as an NamedIndividual. I think it calls back to Class if it can't find anything else, but I need to check.

phillord:
Hmmm. Or not

```
    <owl:Class rdf:about="https://www.enzyme-database.org/query.php?ec=1.1.1.n1">
        <obo:IAO_0100001 rdf:resource="https://www.enzyme-database.org/query.php?ec=1.1.1.305"/>
[...]

PR #132, opened 2025-02-24
io::owx::reader: fixes todo!()

filippodebortoli:
Addresses #129 partially.

@phillord could you take a look and see if this is the intended behavior?


phillord:
I think this all looks good also. The new `matches` macro cleans up those tests very nicely. Please feel free to merge to devel.

Issue #135, opened 2025-03-14
Add horned-validate binary


phillord:
Like horned-parse but returns an exit code to show unsucessful parsing. Or just update horned-parse to do the same.

phillord:
@filippodebortoli confused by PR. Are you working on this?

filippodebortoli:
> [@filippodebortoli](https://github.com/filippodebortoli) confused by PR. Are you working on this?

Yes, but I first misunderstood what you meant.
I will soon push a new PR that should be addressing what you actually wanted.

filippodebortoli:
Closed by #138.

PR #138, opened 2025-03-14
new tool: horned validate.

filippodebortoli:
Addresses #135.
It essentially combines `horned parse` and `horned unparsed`.
If no incomplete parse is present, then the program exits successfully.
If an incomplete parse is found, it will be printed and the program will exit with 1.

- The RDF parser was returning an incomplete parse even if the ontology was parsed successfully, I changed that behavior so that if the parse is complete then there is no incomplete parse returned after decomposing the parser.
- It is very rudimental for now: everything is printed on STDOUT. Ideally, we could print the unfinished part on STDERR.

### DIFF
--- a/horned-bin/Cargo.toml
+++ b/horned-bin/Cargo.toml
@@ -66,3 +66,7 @@ path = "src/bin/horned_triples.rs"
 [[bin]]
 name = "horned-unparsed"
 path = "src/bin/horned_unparsed.rs"
+
+[[bin]]
+name = "horned-validate"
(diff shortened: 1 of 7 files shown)


phillord:
Looks good to me. 

We might want to reconsider or reopen a new issue. #135 is addressed by this, but thinking about the exit codes in the commands that we have probably also makes sense. horned-parse should probably return failure if the parse fails (i.e. unparsed triples. And the other commands could do with a review.

filippodebortoli:
I opened #140 to continue the conversation.

PR #139, opened 2025-03-17
Non deterministic RDF parser

### LINKED ISSUE #123: RDF parser randomly skips components
When trying to parse the OpenEnergyOntology (https://openenergy-platform.org/ontology/oeo) the RDF parser  skips over ~80 components and returns a different set of components each time. I.e. the following test fails and `diff` is different on each run. Unfortunately, I couldn't make out a pattern yet.

```rs
        let file = File::open("oeo.owl").unwrap();
        let mut f = BufReader::new(file);

        let o1 = match read(&mut f, Default::default()) {
            Ok((o, _)) => { SetOntology::<RcStr>::from(o)}
            Err(_) => { assert!(false); panic!() }
        }.into_iter().collect::<HashSet<_>>();

        f.seek(SeekFrom::Start(0));

        let o2 = match read(&mut f, Default::default()) {
            Ok((o, _)) => { SetOntology::<RcStr>::from(o) }
            Err(_) => { assert!(false); panic!() }
        }.into_iter().collect::<HashSet<_>>();

        let mut diff: Vec<_> = o1.difference(&o2).collect();
        diff.sort();

        for x in &diff {
            dbg!(x);
        }

        assert!(diff.is_empty())
```

The error does seem to be related to the OEO; other (large) ontologies in RDF/XML do not produce this error. Also, other parsers don't produce this error.

b-gehrke:
Fixes RDF/XML parser's non-deterministic parsing of class expressions.

Closes #123 .

### DIFF
--- a/src/io/rdf/reader.rs
+++ b/src/io/rdf/reader.rs
@@ -1063,7 +1063,7 @@ impl<'a, A: ForIRI, AA: ForIndex<A>, O: RDFOntology<A, AA>> OntologyParser<'a, A
     }
 
     fn class_expressions(&mut self, ic: &[&O]) -> Result<(), HornedError> {
-        let class_expression_len = self.class_expression.len();
+        let mut parsed_new_ce = false;
         for (this_bnode, v) in std::mem::take(&mut self.bnode) {
             // rustfmt breaks this (putting the triples all on one
             // line) so skip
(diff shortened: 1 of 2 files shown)


b-gehrke:
@phillord do you think this can be merged?

filippodebortoli:
> @phillord do you think this can be merged?

@phillord I would merge this PR unless you have objections to it.

phillord:
Sorry for delay! 10 seconds to unignore the test. 3 minutes to work out why clippy was being silly. 15 minutes to get git to work.

PR #143, opened 2025-04-02
Make bench mark files generated

### LINKED ISSUE #142: Benchmarks files are incomplete or missing.
While running `cargo bench` I noticed that:

- the [food ontology](https://github.com/phillord/horned-owl/blob/devel/benches/ont/food.owl) contains undeclared object properties, e.g. `hasFlavor`, and the parse is incomplete.
- `o10.owl` is missing, and the benchmark is interrupted.

@phillord could it be that some files that are necessary for benchmarking have not been uploaded to the `benches` folder yet? Or am I supposed to run the benchmarks differently, e.g. through `make`? Regarding the food ontology, is the online version outdated or are you using the same file?

phillord:
https://github.com/phillord/horned-owl/issues/142


filippodebortoli (approved):

phillord:
Think this version is better -- no need to use the makefile now at the cost of making things before complex.

filippodebortoli:
@phillord is this PR ready to be merged, or were you planning to add further commits to it?

Issue #144, opened 2025-05-19
Support catalog files


phillord:
Horned-OWL does not use catalog files.

https://github.com/ontology-tools/py-horned-owl/issues/43

filippodebortoli:
@phillord do you already have any idea on how the library should support XML catalogs?
Is this mostly a matter of changing the methods for resolving imports to use an (optional) catalog file, and adding a method to check that the catalog is valid (i.e. all imports in the catalog point to existing files) that can be invoked before parsing?

phillord:
Yes, it is just about handling imports.

Currently, Rust takes the IRI that is imported and resolves it. As #153 shows it also assumes that the IRI that is used for importing is both the physical location and Ontology IRI.

phillord:
Resolved by #224: new \`horned-catalog\` crate parses OASIS XML Catalog (\`catalog-v001.xml\`) files and resolves an IRI to a local path, wired into \`resolve_iri\` as the first resolution step (ahead of the existing path-guessing heuristic and network fallback). Design and scope in \`docs/horned-catalog-plan.md\` on \`devel\`.

PR #146, opened 2025-06-05
resolve: changes return type and signature of localize_iri

filippodebortoli:
This is a small change to the signature and return type of `resolve:localize_iri()`.

It previously required the argument `doc_iri` to be `IRI<A>`, but the actual usage of this method suggests that `doc_iri` can simply be a `str`, avoiding the unnecessary creation of a IRI builder and the allocation of an IRI in `materialize_1`.

Additionally, we can return a `String` (or potentially even a `Path`) instead of an IRI, thus avoiding a second time the creation of an IRI builder for allocating an IRI.

@phillord is there any objection to merging? The idea would be to incrementally rebuild this method to be effective in locating imports also using catalogs as mentioned in #144 and [this issue for py-horned-owl](https://github.com/ontology-tools/py-horned-owl/issues/43).

### DIFF
--- a/horned-bin/src/lib.rs
+++ b/horned-bin/src/lib.rs
@@ -126,12 +126,12 @@ pub fn materialize_1<'a>(
     let amont: RcComponentMappedOntology = parse_imports(Path::new(input), config)?.into();
     let import = amont.i().import();
 
-    let b = Build::new_rc();
+    let _b = Build::new_rc();
 
     // Get all the imports
     for i in import {
         if !done.contains(&i.0) {
-            let local: String = localize_iri(&i.0, &b.iri(input)).into();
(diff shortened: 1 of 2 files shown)


phillord:
I am afraid I didn't see this at the time and it has become outdated by other changes that I have made to the interface of these methods.

I think what you were suggesting would have made sense; but I do quite like the clarity of using `IRI` as a newtype to distinguish between this and a generic `str` type. I have found a number of bugs which have stemmed from thinking something was an `str` representing an `iri` when it is actually a file path.

I wonder about the necessity for creating a `Build` instance. Clearly it is meant as optimisation. It might make sense to remove the privacy restriction on `IRI`, which would enable client code to create a new `IRI` without a `Build` object.

Issue #145, opened 2025-06-02
Make the serialization order consistent with the owlapi


cmungall:
Some time ago the owlapi undertook the herculean effort of making the serialization of an ontology deterministic:

 - https://github.com/owlcs/owlapi/issues/273

This made it much easier to work with OWL in version control, as we have been doing with obo format since 2002. There are still some quirks, it's deterministic given a particular version of the OWLAPI, but increments of the OWLAPI can change things.

Now, I'm all geared up to use (py-)horned-owl in ontology development pipelines, the first thing I realized is that doing so introduces a massive spurious diff...

This is not the end of the world. I can have a pipeline in which I use robot to normalize the serialization. In fact the ODK includes a standard target for just this purpose. But it's a little janky to chain these together in pipelines (and we lose some of the speed benefits of rust).

Making all serializations consistent is likely impossible, I am not even sure there is a spec for what the owlapi does. But if we could prioritize ofn that could be wonderful (ofn and obo are the only sane choices for the edit version of an ontology).

phillord:
I know this is a problem. I first tried putting OWL generated by protege into a VCS more than two decades ago and it was utterly impossible.

I *believe* that the serialisation order for horned-owl is deterministic -- I test this implicitly in the unit tests, although most of them are quite short so only have a single or small numbers of axioms of few different types. I thing that if this does not happen, it would be a bug.

Now, deeper than this, well, that is much harder, particularly if we take it to the levels of whitespace. For the RDF/XML representation, the actual output is dependent on the OWL part of horned, the prettification step in pretty_rdf and finally the XML level output of quick_xml. 

The OFN might be a bit more plausible. It would be easy to write a test that does a whitespace diff on roundtripped OFN. It then all depends on how sane the OWL API output is. If it is written like @althonos's renderer, with most of the white space layout defined in a few places, it might work well.

cmungall:
Thanks @phillord 

Ah yes I should have been more precise - deterministic is good but not enough for mixed protege/horned-owl workflows. We need horned-owl to implement an unwritten canonical order serialization specification. And yes, one that it down to the level of whitespace and comments.

As far as I can tell the main differences for ofn rendering are:

- owlapi does not indent the axioms with whitespace
- owlapi ordering is more frame-oriented:
    - all declarations first
    - Annnotation Properties before Object Properties before Classes
    - Each entity type has a block comment
    - Each entity has a fresh comment that includes the label `Class: obo:CL_0000000 (cell)`
[...]

PR #147, opened 2025-06-06
DRAFT: Feature/owl api equivalent ofn

phillord:


phillord:
https://github.com/phillord/horned-owl/issues/145

phillord:
I have the framework in place for this now. It doesn't look too hard actually.

@althonos Do you care about the current indentation of the OFN? We can do this as parser configuration so that we can either use your current format or an OWL API compliant one. Or we can just have a single spacing output which is OWL API complaint. I'd prefer the latter, just because it's easier.

althonos:
@phillord : I'm happy with any change, in particular when that makes it more compliant to the OWL API!

phillord:
Will close this. May come back to it at a later date

Issue #151, opened 2025-06-27
OWX default prefix uses ontologyIRI instead of the empty prefix


janekx21:
In the reader for OWX files the default prefix is set to the `ontologyIRI` attribute of the ontology tag but the empty prefix eg. `<Prefix name="" IRI="http://some/url#"/>` should be used.

```xml
<?xml version="1.0"?>
<Ontology xmlns="http://www.w3.org/2002/07/owl#"
     xml:base="http://www.semanticweb.org/janek/ontologies/2025/5/untitled-ontology-3"
     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
     xmlns:xml="http://www.w3.org/XML/1998/namespace"
     xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
     ontologyIRI="http://www.semanticweb.org/janek/ontologies/2025/5/untitled-ontology-3">
    <Prefix name="" IRI="http://www.semanticweb.org/janek/ontologies/2025/5/untitled-ontology-3/"/>
    <Prefix name="owl" IRI="http://www.w3.org/2002/07/owl#"/>
    <Prefix name="rdf" IRI="http://www.w3.org/1999/02/22-rdf-syntax-ns#"/>
    <Prefix name="xml" IRI="http://www.w3.org/XML/1998/namespace"/>
    <Prefix name="xsd" IRI="http://www.w3.org/2001/XMLSchema#"/>
    <Prefix name="rdfs" IRI="http://www.w3.org/2000/01/rdf-schema#"/>
</Ontology>
```

The correct default value is `http://www.semanticweb.org/janek/ontologies/2025/5/untitled-ontology-3/` with a slash at the end. This is important, because the ontology IRI can differ from the empty prefix (The use case for this is importing an ontology).

This causes incorrect full IRI resolution.

Ref:
https://github.com/phillord/horned-owl/blob/93277bfbc3100066e2dea941864a1125268b9af3/src/io/owx/reader.rs#L63

filippodebortoli:
@janekx21 the problem should lie here:

https://github.com/phillord/horned-owl/blob/93277bfbc3100066e2dea941864a1125268b9af3/src/io/owx/reader.rs#L71C1-L85C22

Line 76 should be replaced with something along the lines of:

```rust
if p.is_empty() {
  r.mapping.set_default(&i);
} else {
[...]

PR #152, opened 2025-07-04
owx/reader: fixes handling of empty (default) prefix.

filippodebortoli:
Addresses #151.


phillord:
This one has hung around for a while (my fault!), so will merge now

Issue #153, opened 2025-09-01
horned-parse panic at closure_reader.rs:127:37


KonradHoeffner:
Using horned-parse 1.1.0 as installed with `cargo install` I get a panic with several horned tools including horned-parse:

```
$ horned-parse annodc.owl    

thread 'main' (4416) panicked at /home/konrad/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/horned-owl-1.1.0/src/io/rdf/closure_reader.rs:127:37:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Attached is annodc.owl zipped : [annodc.zip](https://github.com/user-attachments/files/22078960/annodc.zip)

phillord-ncl:
This is the import resolution mechanism. 

Changing 

````
<owl:imports rdf:resource="https://w3id.org/gfo-light/release/2025-03-17"/>
````

to 

`````
        <owl:imports rdf:resource="https://w3id.org/gfo-light"/>
`````

and saving the import as `gfo-light.owl` now makes everything work.

I will have a think on how to fix this.

KonradHoeffner:
Unfortunately applying your suggested change and downloading the file in that directory doesn't solve the issue for me because now it fails at the same place in the code due to gfo-light.owl importing <https://w3id.org/gfo-core/release/2025-04-07> in a similar manner.

filippodebortoli:
This is possibly related to #144 and to the idea of refactoring the module to support XML catalogs.

phillord:
Closing as fixed. \`ClosureOntologyParser\` now keeps an alias map from an ontology's plain IRI to its version IRI (\`src/io/rdf/closure_reader.rs\`), so an \`owl:imports\` statement referencing the plain IRI resolves correctly even when the imported ontology also declares a version IRI — exactly the annodc.owl → gfo-light → gfo-core chain from this issue.

Verified by actually reproducing it: built \`horned-parse\` from \`devel\` and ran it against the original \`annodc.owl\` attachment (with the import chain resolved over the network, not from a local cache) — parses cleanly with no panic.

PR #154, opened 2025-09-02
add panic message when ontology URI cannot be imported

KonradHoeffner:
When an ontology cannot be imported the user currently just gets a panic without explanation which makes troubleshooting very difficult.
As a first step of solving issue #153, this PR adds a more helpful message.


phillord-ncl:
We should be returning an error here rather than panic'ing. 

closure_reader needs a bit of love! It's got quite a few `unwrap`s and and a `todo`

phillord-ncl:
I will have a bash at it and see if I can improve things

phillord:
Will close now, as we have better error handling

PR #156, opened 2025-09-02
fix GitHub action check by fixing clippy warnings

KonradHoeffner:
All new pull requests currently fail because the recent commit hook check fails at a single Clippy warning.
This PR applies both the automatic Clippy fixes and also allows OrTerm to be unused so that the PR system can continue functioning.


filippodebortoli:
@phillord I would merge this soon. Any objection?

filippodebortoli:
@KonradHoeffner @phillord do these still need to be fixed after #158? If so, I would merge today.

KonradHoeffner:
@filippodebortoli good point, I just rebased and there were no changes, so I removed this.

Issue #155, opened 2025-09-02
upgrade to Rust 2024 edition


KonradHoeffner:
I started learning Rust with the 2021 edition so contributing to this library which still uses the 2018 is difficult, maybe other contributors face this problem as well.
Would you be open to upgrade to the current 2024 edition or is that not planned?
I could also try to do it.

filippodebortoli:
Closed by #158.

PR #158, opened 2025-09-02
upgrade Rust edition from 2018 to 2024

KonradHoeffner:
I used `cargo fix --edition` and then upgraded the "edition" field in `Cargo.toml` and `horned-bin/Cargo.toml` to upgrade from 2018 to 2021 edition and then repeated the process to upgrade from 2021 to 2024. This PR resolves issue #155.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ categories = ["command-line-utilities", "parsing",
 
 license = "LGPL-3.0"
 edition = "2024"
+rust-version = "1.88"
 
 [dependencies]
 curie = "0.1.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ categories = ["command-line-utilities", "parsing",
               "rendering", "science", "data-structures"]
 
 license = "LGPL-3.0"
-edition = "2018"
+edition = "2024"
 
 [dependencies]
 curie = "0.1.1"
--- a/benches/io.rs
+++ b/benches/io.rs
@@ -12,7 +12,7 @@ fn io_read(c: &mut Criterion) {
     for n in [10, 100, 1_000, 2500, 5000, 10_000].iter() {
         group.bench_with_input(BenchmarkId::new("owl_io_read", n), n, |b, &n| {
             b.iter(|| {
-                let f = File::open(format!("benches/ont/o{}.owl", n)).ok().unwrap();
+                let f = File::open(format!("benches/ont/o{n}.owl")).ok().unwrap();
                 let mut f = BufReader::new(f);
                 let _ = horned_owl::io::rdf::reader::read(&mut f, Default::default()).ok();
             })
@@ -20,7 +20,7 @@ fn io_read(c: &mut Criterion) {
 
         group.bench_with_input(BenchmarkId::new("owx_io_read", n), n, |b, &n| {
             b.iter(|| {
-                let f = File::open(format!("benches/ont/o{}.owx", n)).ok().unwrap();
+                let f = File::open(format!("benches/ont/o{n}.owx")).ok().unwrap();
                 let mut f = BufReader::new(f);
                 let _: (SetOntology<RcStr>, _) =
                     horned_owl::io::owx::reader::read(&mut f, Default::default())
--- a/benches/io.rs
+++ b/benches/io.rs
@@ -1,4 +1,4 @@
-use criterion::{criterion_group, AxisScale, BenchmarkId, Criterion, PlotConfiguration};
+use criterion::{AxisScale, BenchmarkId, Criterion, PlotConfiguration, criterion_group};
 use horned_owl::model::RcStr;
 use horned_owl::ontology::set::SetOntology;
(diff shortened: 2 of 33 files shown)


phillord-ncl:
Bit irritating that `cargo fix` seems to create so many diffs that are not about edition changes. As far as I can see that relates to pretty much all of these changes!

The exception is the use of the `expr_2021` fragment specifier. I think cargo is being conservative here. It's cannot work out the intention nor understand whether the 2024 version of `expr` is correct so is picking the `expr_2021` for backward compatibility. I can't work it out for sure either, because I haven't made a lot of use of `const` expressions. Overall, though, I think we should use the `expr` fragment; clearly it's going to be the future and that will be how people expect it to behave.

It also compiles file and tests so it must be good!

phillord-ncl:
I think, also, we should just kill `OrTerm` off. There is no point having dead code around.

There are also two cases of `use TryInfo/TryFrom`. Bit surprised that clippy doesn't pick those up, but they are not needed in Rust 2024 (but where in 2018).

KonradHoeffner:
I've noticed that there are a lot of new lints coming out all the time so I guess that an upgrade of two editions at once would produce a lot of warnings.

I don't know if there is an option to fix errors only but since the current CI pipeline rejects on warnings anyways I think it is working as intended by making use of new features of the 2024 such as implicit argument capture in the panic macro.

There are also a lot of import reorderings from the [2024 edition Rustfmt version sorting change](https://doc.rust-lang.org/edition-guide/rust-2024/rustfmt-version-sorting.html) and as the CI pipeline also checks Rustfmt I think this is necessary as well.

Clippy also now uses let chains such as those:

```
-        if let Some(ref mut it) = self.inner {
-            if let Some(component) = it.next() {
-                return Some(component.borrow());
-            }
+        if let Some(ref mut it) = self.inner
+            && let Some(component) = it.next()
+        {
+            return Some(component.borrow());
         }
```

This only got stabilized in Rust version 1.88 onwards so I tried to set `rust-version = "1.85" in Cargo.toml, however that does not cause clippy to downgrade the code (it seems to only work forwards, not backwards).
Thus I changed the minimum supported Rust version to 1.88 and hope that is OK with you.
If not and you intend an earlier version like 1.85 then I can manually downgrade all those places where if-let chains are used.
I verified that it compiles and tests successfully with 1.88, only clippy has warnings with 1.88 that do not occur with newest stable and newest nightly.
But because the CI pipeline used in the GitHub action seems to used newest (stable or nightly? both are OK), it all seems to work out in the end.

phillord-ncl:
Yeah, there are a few other changes that I am surprised about. The extra space in:

````
Functional(self.0 .0, self.1, None),
````

has been removed. I am sure that it was clippy that put these in as it seems an unnatural spacing to me anyway.

Also it has refactored this:

````
        if let Some(r) = rest.take() {
            triples!(f, bn.clone(), ng.nn(RDF::Rest), r);
        } else {
            triples!(f, bn.clone(), ng.nn(RDF::Rest), ng.nn(RDF::Nil));
````

into a `match` statement which again I am surprised about.

I am largely happy to update to 1.88 though. Horned is still more research software than heavy deployment base. We can wait for a day or two to see if anyone complains or even revert later if we really needed to.

filippodebortoli:
@phillord-ncl I can take care of the merge if there is no other reaction within the next 24 hours.

phillord:
@filippodebortoli Need to fix the "expr_2021"  and remove the undeed TryInto, TryFrom imports first. Once that has been done, we can do a final review before merge

KonradHoeffner:
By the way I noticed that it doesn't compile when having "--all-features" active but that is true for the baseline "devel" branch as well, ist that intended?

KonradHoeffner:
The unneeded TryInto and TryFrom imports are now removed.

P.S.: There is a rustc lint allowed by default called [redundant_imports](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#redundant-imports), I found more of them using `cargo clippy -- -W redundant_imports` and removed them as well in a separate commit.

KonradHoeffner:
OK so I made the change from expr_2021 back to expr in commit f603492 and the tests are successful but I have no experience with this expr statement at all, so if there are things that are not covered by the automatic test suite I can't guarantee for that having equal functionality.

phillord:
> By the way I noticed that it doesn't compile when having "--all-features" active but that is true for the baseline "devel" branch as well, ist that intended?

No, that is a bug. It's the "encoding" feature that is guilty; it is something that needs fixing at some point, probably by removing the feature.

phillord:
The `expr` fragment just identifies `const` expressions which it didn't use to because they didn't exist. It's a reasonable guess that accepting this change is not going to cause any problems.

phillord:
I think this looks good to go and ready. Overall, I think that the use of the new features, particular the `{format}` variable interpolation makes the code cleaner!

phillord:
Many thanks for the work!

KonradHoeffner:
You're welcome! A lot of the work was done by Clippy :-)

PR #159, opened 2025-09-03
Clean up and better error handling in ClosureOntologyParser

phillord:
This is my attempt to improve error reporting.

https://github.com/phillord/horned-owl/pull/154
https://github.com/phillord/horned-owl/issues/153

I am still wondering whether there is a way we can get rid of the `unsafe` in closure_reader.rs. I think that there should be a way using `RefCell`, but it might make quite a bit of the code more complex.


phillord:
Will merge this. Will worry about unsafe at another point. I think it is pretty hard to remove.

PR #166, opened 2025-10-06
adds getters for io::rdf::reader::PosTriple

### LINKED ISSUE #165: Can not access PosTriple inner data
https://github.com/phillord/horned-owl/blob/93277bfbc3100066e2dea941864a1125268b9af3/src/io/rdf/reader.rs#L386

The struct PosTriple is public but I cant find a way to access the inner data. Incomplete RDF parses are therefore unusable.

Image

filippodebortoli:
Resolves #165.

Adding getters for `PosTriple`, both for the triple (ref, mut ref, consuming) and for the position (ref).

### DIFF
--- a/src/io/rdf/reader.rs
+++ b/src/io/rdf/reader.rs
@@ -396,6 +396,24 @@ impl<A: ForIRI> From<[Term<A>; 3]> for PosTriple<A> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


phillord:
Thanks you for this, good solution!

I am hoping no one pulled my original merge, because I messed up the PR and had to remove everything with a force push to devel.

Issue #164, opened 2025-10-06
RDF parser panics with foaf


janekx21:
```shell
$ curl -H "Accept: application/rdf+xml" http://xmlns.com/foaf/0.1/ -L > /tmp/foaf.owl
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 44209  100 44209    0     0   171k      0 --:--:-- --:--:-- --:--:--  172k
$ horned summary /tmp/foaf.owl

thread 'main' panicked at /home/janek/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/horned-owl-1.1.0/src/io/rdf/reader.rs:1541:34:
not yet implemented
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This panics at a todo macro. As a workaround I use https://github.com/pchampin/sophia_rs now.

https://github.com/phillord/horned-owl/blob/93277bfbc3100066e2dea941864a1125268b9af3/src/io/rdf/reader.rs#L1541C34-L1541C38
```rust
[pr, Term::RDF(VRDF::Type), Term::OWL(VOWL::InverseFunctionalProperty)] => {
    ok_some! {
        match self.find_property_kind(pr, ic)? {
            PropertyExpression::ObjectPropertyExpression(ope) => {
                InverseFunctionalObjectProperty(ope).into()
            },

            _ => todo!()
        }
    }
}
```

phillord:
That's an interesting one. FOAF declares `mbox_sha1sum` to be a `DataProperty` but then also says that it is `InverseFunctional`. Which doesn't fit into the OWL data model directly.

The comments in the RDF file suggest that they knew what they were doing when they did this.

I will think on a solution.

janekx21:
Hay thanks for the quick awnser.

janekx21:
I would like to see all todo! statements removed. A lib should not panic on parsing error.

phillord:
I'd agree -- that's why they are "todo" rather than "panic!". I'll update this in a few days times. It's made me rethink how much I want to validate the OWL and how much I just want to parse it in the RDF parser. I think it is doing too much validation at the moment. Reducing this also reduces the amount of error handling I need to do which also means that todo! statements get removed.

phillord:
And complete!

Thanks for the original bug report

PR #167, opened 2025-10-07
Foaf reading

phillord:
Starts to address #164


phillord:
With this version FOAF now parses without error, but does not parse completely.

This has involved a fairly radical overhaul of the RDF parser, meaning that it does not check the declared type of an entity where it can infer what it should be, which is invalid in some profiles of OWL2.

phillord:
This now parses all but four triples in foaf. I am still trying to work out how we know the type of "isPrimaryTypeOf"

phillord:
This now parsers foaf without error in lax mode. It has to use unsupported guesswork to do so as FOAF is not fully specified as an ontology. Some clean ups needed, but comments welcome.


Fixing this has resulted in a rather cleaner RDF parser, so I am very glad for the original bug report!

PR #171, opened 2025-11-05
Add default prefix to prefix map

### LINKED ISSUE #151: OWX default prefix uses ontologyIRI instead of the empty prefix

b-gehrke:
Closes #151

### DIFF
--- a/src/io/owx/reader.rs
+++ b/src/io/owx/reader.rs
@@ -72,10 +72,9 @@ pub fn read_with_build<A: ForIRI, O: MutableOntology<A> + Default, R: BufRead>(
                         let prefix = get_attr_value_str(&mut r.reader, e, b"name")?;
                         match (prefix, iri) {
                             (Some(p), Some(i)) => {
+                                let _ = r.mapping.add_prefix(&p, &i);
                                 if p.is_empty() {
                                     r.mapping.set_default(&i);
-                                } else {
-                                    let _ = r.mapping.add_prefix(&p, &i);
                                 }
                             }
                             (None, _) => {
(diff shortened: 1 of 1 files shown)


phillord:
Many thanks!

PR #173, opened 2026-06-05
Parse axiom annotations on SWRL rules in RDF

balhoff:
Annotated SWRL rules were silently dropped when read from RDF syntaxes.

The RDF reader matched a rule's `swrl:Imp` node against an exact three-triple slice (`type`/`body`/`head`); an annotation on the rule (e.g. `rdfs:label`) adds a fourth triple, the match failed, and the entire rule was discarded. OWL Functional Syntax was unaffected.

The OWL API (and Protégé) serialize rule annotations as direct triples on the Imp node, not as reified `owl:Axiom`.


phillord:
Thank you!

PR #174, opened 2026-06-10
Manchester Syntax Parser

b-gehrke:
Continuation of @althonos work in #103

Adds:
- Manchester Syntax Extensions (complex class expressions in class frame targets)
- OMN test files generated with OWL API 5 (which supports complex class expressions in class frame targets)
- Ambiguity resolving
  - record entity types during parsing
  - parse ambiguous components as the first matching but store for later processing
  - process all ambiguous components and convert if needed
- Integration in horned-bin

One test is skipped because the OWL API does not create all axioms for MS (`gci_and_other_class_relations.clj`)


phillord:
Oh dear, now we have two!

https://github.com/phillord/horned-owl/pull/176

I would be happy to hear thoughts about whether we want one or two or a combination of the two.

b-gehrke:
Closing in favour of #176 as discussed in #179.

Issue #175, opened 2026-06-14
Annotation model has no slot for nested annotations (annotation-on-annotation silently dropped by all readers)


micheldumontier:
## Summary

`horned_owl::model::Annotation { ap, av }` has no field to hold annotations **on the annotation itself**. OWL 2 permits annotated annotations, and both standard concrete syntaxes express them:

- **Functional Syntax:** `Annotation( Annotation( :m "why" ) rdfs:comment "hello" )`
- **Manchester Syntax** (`annotationAnnotatedList`): `Annotations: Annotations: :m "why" rdfs:comment "hello"`

Annotations on **axioms** are supported via `AnnotatedComponent.ann`, but an annotation nested on an `Annotation` value has nowhere to be stored.

## Impact

Every reader (`ofn`, `owx`, `rdf`, and the new `omn`) parses the nesting and then **silently discards the inner annotation** — the outer annotation survives, its own annotation is lost on round-trip. This is silent information loss for ontologies that use annotated annotations.

## Repro (Functional Syntax)

```
AnnotationAssertion( Annotation( :m "why" ) rdfs:comment :A "hello" )
```

round-trips as

```
AnnotationAssertion( rdfs:comment :A "hello" )
```

— the `Annotation( :m "why" )` is gone.

## Possible direction

Add an `ann: BTreeSet<Annotation<A>>` field to `Annotation` (mirroring `AnnotatedComponent`), threaded through the reader/writer for each syntax. This is a cascading model change touching every io path, so it likely warrants its own design discussion.

## Context

Surfaced while adding an OWL 2 Manchester Syntax (`io::omn`) reader+writer. The new OMN reader deliberately matches the existing OFN reader's parse-and-drop behavior here, so this is a **pre-existing model limitation**, not specific to the Manchester work.

phillord:
The joys of OWL. There's always an edge case that you have forgotten. On a positive side, I know that Tawny-OWL supports this because I forgot it when I wrote that as well; so at least I will be able to generate some test cases.

johanwk:
Indirectly related -- with [ELOT](https://github.com/johanwk/elot) annotations on annotations are very easy to use (indented sub-list), and therefore they are also frequently used. When Horned-OWL get support for Manchester Syntax, ELOT will appreciate it very much :)

phillord:
[...]

PR #176, opened 2026-06-14
Add OWL 2 Manchester Syntax (OMN) reader + writer (io::omn)

micheldumontier:
## Summary

Adds `io::omn` — a general **OWL 2 Manchester Syntax** reader and writer. horned-owl previously shipped `ofn` / `owx` / `rdf` io only; this fills in the Manchester gap (reader + writer + CLI wiring).

The grammar targets the full W3C OWL 2 Manchester Syntax §2.5: prefix/ontology header (incl. version IRI + imports), all six entity frames, datatype definitions, full data ranges (`and`/`or`/`not`/`{…}`/facets), all literal forms, the misc axiom section, per-item annotated lists, anonymous individuals, and general class axioms via `Class: <expr>` frames.

## Validation

Conformance is checked against the **OWL-API (ROBOT) oracle** plus a per-construct matrix, in `tests/manchester/` (a report generator writes `docs/manchester/compliance-report.md`):

- **All four corpus ontologies parse *and* round-trip:** koala, sio, obi-core, hp (346k components).
- **103 §2.5 constructs** read + write + round-trip clean; **5 documented residuals**, each asserted in a compiler-exhaustive match (no rubber-stamp):
  - SWRL `Rule:` (no §2.5 syntax),
  - bare default-prefix local names (not lexable),
  - complex-LHS GCIs emitted into a `# General axioms` functional block (our writer now uses native `Class: <expr>` frames instead, but a foreign functional block is skipped-with-warning),
  - **nested annotation-on-annotation — a pre-existing model limitation, see #175**,
  - the undeclared-property object/data tail (a property/filler never declared in the document defaults to object; declared cases are resolved from `DataProperty:`/`Datatype:` frames).
- **GCI handling is OWL-API-consistent both ways:** we parse ROBOT's `Class: <complexExpr>` GCI frames, and ROBOT re-parses our GCI output back to the same axiom.
- No-panic fuzz (4000 cases) + adversarial fixtures (unicode, deep nesting, CRLF, escaping).

## CLI

`horned`, `horned-dump`, and `horned-round` now handle `.omn` (read + render), via `ResourceType::OMN` / `ParserOutput::omn`.

## Notes

- Rebased onto `main` (1.4.0). The history is granular (incremental development + conformance fixes) — happy to **squash** if you'd prefer a single commit.
- Object-vs-data property ambiguity (Manchester can't lexically distinguish them in `HasKey:`, restrictions, and `Equivalent/DisjointProperties:`) is resolved by an order-independent declaration pre-pass over the buffered frames, sound by OWL 2 DL object/data disjointness.


micheldumontier:
Force-pushed: rebased onto current `devel` (which now carries the nested-annotation `ann` field from #178) and added one commit wiring nested annotations through the OMN reader/writer — the OMN entry's leading `Annotations:` is now stored in `Annotation.ann` and re-emitted as `Annotations: <nested> ap av`, matching the OFN/OWX/RDF behaviour you landed. Also carries two small OMN clippy fixes.

Verified locally against the rebased `devel`: `cargo test` (1017 lib + manchester conformance + integration), `cargo clippy -- -D warnings`, and `cargo bench --no-run` all green.

phillord:
Ready?

micheldumontier:
I am running against bioportal ontologies to make sure they round-trip.

micheldumontier:
Pushed one commit adding two round-trip fidelity fixes found by running the reader/writer over the BioPortal corpus:

- **(C)** literal-enumeration data restrictions (`p only {"a","b"}`) now parse as `DataOneOf` instead of failing with "expected Individual".
- **(D)** the `# General axioms` fallback block is now read back via the functional-syntax reader rather than skipped — it was silently dropping axioms (one ontology lost all ~38k annotation assertions on round-trip) — with a safe warn-and-skip fallback if the block can't be parsed.

Full lib + conformance suites green. Two shared OFN-writer bugs this surfaced (multibyte-literal escaping, blank-node prefix) are in #183.

— by Claude with @micheldumontier

micheldumontier:
@phillord i think we are good to go.

b-gehrke:
Just out of curiosity: What kind of axioms have to be written in Functional Syntax in the "General axioms" block? With the extension of the class frame and object frame to support complex class/object expressions all of OWL should be able to be expressed in OWL.

micheldumontier:
Two more commits closing the object-vs-data-property restriction ambiguity (the last gap the BioPortal round-trip surfaced):

- **Grammar**: the data-arm lookahead now sees through a leading `not`, so a negated data-shaped filler (`dp some not xsd:float[…]`) is read as a `DataComplementOf` instead of hard-erroring on the facet bracket (this was PACO) or silently mis-binding to an object complement.
- **Declaration flip**: a negated *custom* datatype known only by declaration (`p some not :MyType` with `Datatype: :MyType`, or `p` a declared `DataProperty:`) now flips to the data variant over `DataComplementOf` across all five restriction keywords (`some`/`only`/`min`/`max`/`exactly`).

An undeclared bare IRI filler remains object-form — irreducibly ambiguous without a declaration, as it is for the OWL API. Full lib + conformance suites green.

— by Claude with @micheldumontier

phillord:
Merged and many thanks!

micheldumontier:
Heads-up: the `pre-commit` CI failure was rustfmt (`cargo fmt --check`) — my restriction-handling code needed rustfmt's line wrapping, and it also surfaced some pre-existing formatting drift in `io/omn/writer/mod.rs`. Fixed in `26566ee` (`cargo fmt`, no behavioural change; full lib suite still green). The branch is now fmt-clean, so it just needs a CI re-run when you get a chance.

— by Claude with @micheldumontier

micheldumontier:
Good question. In practice — across ~60 BioPortal ontologies, ~2,900 block lines — the only thing that lands there is **`AnnotationAssertion`**, and specifically annotation assertions whose **subject appears *only* as an annotation subject**: no `Declaration`, and not used in any logical axiom. So there's no entity frame to hang the `Annotations:` on.

The writer *could* emit a `Class:`/`Individual:` frame for such an IRI, but our reader (like the OWL-API's) treats a frame as implying a declaration — so that would add a `Declaration` axiom that isn't in the source and change the signature on round-trip. Emitting these as functional-syntax `AnnotationAssertion(...)` keeps the axiom set exact.

You're right that the logical side is fully covered by the (now complex) class/object frames — this residual isn't a logical-expressibility gap. `HasKey` / `DatatypeDefinition` / SWRL rules would also be eligible for the fallback, but SWRL now has a native `Rule:` frame and the other two are frame-expressible; none showed up in the corpus. So the block is, in practice, "annotations on entities the ontology references but never declares."

— by Claude with @micheldumontier

phillord:
BTW, do you have code for roundtripping against bioportal. I was steadily working my way through all the ontologies there but gave up a long time ago. Have you done this systematically?

micheldumontier:
Yes — I ran it systematically over the whole BioPortal OWL-language corpus (~880 RDF/XML/Turtle/OWL-XML/functional files, everything except OBO and the compressed archives). The harness reads each ontology, round-trips it through **all four** concrete syntaxes (ofn/omn/owx/rdf), normalises + compares for fidelity, and records per-syntax read/write timing.

A few things fell out of it:
- It drove three fixes already in flight: the OFN literal/blank-node bugs (#183), the OMN writer's O(n²) frame-assembly blow-up → linear (#186), and the SWRL built-in full-IRI fix (#189).
- **Dispatch has to be by *content*, not extension** — e.g. ~119 files named `.owl` are actually OWL/XML, and a bunch of `.rdf` are Turtle. Extension-based loading is exactly where a systematic pass "gives up." A tiny content-sniffer fixed that and recovered ~150 files that otherwise fail to read.
- Round-trip fidelity + timing per syntax (OWX reads fastest, OMN most compact, etc.).

It's currently scratch code with hardcoded paths, so I'd like to package it properly. How would you most want it?
1. a **standalone repo** you can clone (depends on horned-owl as a crate), or
2. **upstreamed** as a `horned-bin` subcommand + a small `dev/` download script (BioPortal API key + target dir, stores files gzipped), or
3. just the **methodology + results + download script**, no round-trip harness.

One related question: horned-owl doesn't currently sniff gzip (readers take raw `BufRead`). If we store the corpus gzipped, either the bench decompresses on read, or — if useful to you — I could add transparent gzip sniffing to the readers as a small separate PR. Preference?

— by Claude with @micheldumontier

phillord:
I'd prefer a standalone repo I think. It will just bulk up the horned-owl repo rather. Does it include the bioportal ontologies or just the code to pull them down.

The content based parsing would be useful to have directly in horned. I know that the OWL API does this by just testing parsers till one succeeds, but I would be happy to have a heurstic map between content and format. We can decide what to do with "parse_path" function in horned-bin which uses extension mapping at a later date.

Transparent gzip handling. That would bring in another dependency (like the gzip crate) I think. Clients of horned-owl can use BufReader around a gunziper can it not?

micheldumontier:
Standalone repo is up: **https://github.com/micheldumontier/horned-roundtrip**

To your question — it's **just the pull-down code, not the ontologies** (size + mixed BioPortal licensing). It's an all-Rust `fetch → run → report` harness that depends on horned-owl (git, pinned rev): `fetch` pulls submissions via the BioPortal API (needs your own key) and stores them gzipped; `run` reads each file, writes it to every other format, reads it back, and diffs the models; `report` emits CSV/JSON/Markdown. `run` also takes any local directory, so a corpus is bring-your-own. Format is detected by **content**, not extension (BioPortal extensions lie constantly — plenty of Turtle behind `.owl`).

I ran it over the local BioPortal corpus (1206 files, ~83% parseable; the rest are mostly OBO, which horned-owl has no reader for). It surfaced some real IO defects worth a look — happy to file separate issues with minimal repros for any of these:

- **OFN writer emits functional syntax the OFN reader then rejects** — ~230 ontologies fail write→reread with a consistent `Parsing Error --> N:5`. Looks like a single writer/reader disagreement on one construct; probably the highest-value fix.
- **OWL/XML round-trips poorly** — `owx→owx` self-round-trip only ~39% clean; includes the empty-prefix bug (`Prefix(:=<iri>)` → the writer emits an invalid `xmlns:=""` attribute that won't re-parse).
- **RDF/XML writer panics on ~60 ontologies.**
- **RDF/XML reader reports an incomplete parse on ~49% of RDF/XML inputs** (its own `IncompleteParse` is non-empty — triples silently dropped).
- ~14k axiom-annotation differences across the corpus (annotations dropped/changed on round-trip).

Numbers and per-ontology detail are reproducible from the repo. Want me to open issues for the OFN-writer and OWX ones first?

— by Claude with @micheldumontier

phillord:
I will test this also. Happy to see confirmed bugs particularly where the cause is clear enough that it is clear that it is not a duplicate. I'll look at the content based parse as well as pull that back into horned.

PR #177, opened 2026-06-18
Fix clippy 1.96 unnecessary_fold lint in IRIMappedIndex::index_remove

micheldumontier:
clippy 1.96 (`-D warnings`) flags the `.fold(false, |v, iri| ... || v)` in `IRIMappedIndex::index_remove` (`src/ontology/iri_mapped.rs`) under `clippy::unnecessary_fold`. It's the only clippy lint on `devel` at this toolchain — a fresh `cargo clippy -- -D warnings` is clean after this one change.

### The subtlety
The obvious `.any()` rewrite clippy nudges toward would be **wrong here**: the closure has a side effect (it removes `cmp` from *every* IRI's set), and `.any()` short-circuits on the first `true` — so removals from the remaining sets would be skipped. This replaces the fold with an explicit loop that preserves full evaluation and still returns whether any set actually changed.

```rust
let mut removed = false;
for iri in Self::iris_from_component(cmp).iter() {
    removed = self.mut_set_for_iri(iri).remove(cmp) || removed;
}
removed
```

`cargo test --lib` → 947 passed, 0 failed.

Surfaced while getting the Manchester-syntax work (#176) through the pre-commit clippy gate — this lint pre-exists on `devel`, independent of that PR, so splitting it out here.

### DIFF
--- a/horned-bin/src/bin/horned_round.rs
+++ b/horned-bin/src/bin/horned_round.rs
@@ -44,7 +44,7 @@ pub(crate) fn matcher(matches: &ArgMatches) -> Result<(), HornedError> {
     match res {
         horned_owl::io::ParserOutput::OFNParser(so, pm) => {
             let amo: RcComponentMappedOntology = so.into();
-            horned_owl::io::owx::writer::write(stdout(), &amo, Some(&pm))
+            horned_owl::io::ofn::writer::write(stdout(), &amo, Some(&pm))
         }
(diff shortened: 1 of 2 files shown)


micheldumontier:
Closing — the `clippy::unnecessary_fold` lint this fixes no longer fires on current clippy (1.96) and isn't flagged by CI on `devel` (it surfaced on a fork CI run on 2026-06-18 but not since). Not worth the noise; happy to reopen if it resurfaces.

PR #178, opened 2026-06-22
Add support for nested annotations on annotations (annotationAnnotati…

phillord:
@micheldumontier Hope this works for you. Probably much the same as your solution although mine uses bubo generated autotest files.

If you are happy I will merge to devel.


micheldumontier:
Yes, happy with this — please merge. 👍

It lines up with the approach in my branch and goes further where it counts: full OWX nested read and the RDF `owl:Annotation` reification (reader two-pass + writer), which I'd deliberately left as follow-ups. The two-pass RDF reader matching `owl:Annotation` bnodes back to their parent via annotatedSource/Property/Target is exactly right, and routing the mutable visitor through `annotation` (rather than `annotation_vec`) neatly avoids the unbounded-recursion trap. The bubo-generated fixtures across all formats are a nice touch.

I'll close my PR as superseded by this.

One heads-up for sequencing: this adds the `ann` field to `Annotation`, so once it lands on `devel` the Manchester PR (#176) will stop compiling — its OMN reader/writer construct `Annotation { ap, av }` in a number of places. I'll thread `ann` through those and wire the OMN `Annotations: Annotations:` form to match this PR's coverage, so #176 is ready to rebase straight onto the updated `devel`.

phillord:
rebased on side, so closing here

PR #183, opened 2026-07-03
fix(io/ofn): three literal/individual round-trip bugs (multi-byte escape, script langtags, blank-node prefix)

micheldumontier:
*(Claude Code here, working on Michel's fork; he reviewed before this went up.)*

Three independent bugs in the functional-syntax reader/writer on `devel`, all surfaced by round-tripping a large real-world OWL corpus (BioPortal). Each is small and has a regression test.

**1. `quote()` panics on multi-byte string literals** (`io/ofn/writer/as_functional.rs`)

The escaper found the next `"`/`\` with `chars().enumerate()` (a *character* index) but sliced with `&s[..i]` (a *byte* index). Any literal with a multi-byte character before an escaped `"` or `\` sliced inside a UTF-8 sequence and panicked (`byte index N is not a char boundary`). Fixed by using `char_indices()`; the matched delimiters are single-byte ASCII, so `i + 1` stays a valid boundary.

**2. Language tags with a script subtag fail to parse** (`grammars/bcp47.pest`)

`BCP47_ExtLang` (exactly 3 alpha) greedily consumed the first 3 letters of a following 4-alpha script subtag: `zh-Hans` parsed as `zh` + extlang `Han`, stranding `s`, so the enclosing literal failed. Region subtags (`en-US`) were unaffected. Added `!ASCII_ALPHA` guards so an extlang subtag yields when a fourth letter follows, letting the script subtag match. `zh-Hans` / `sr-Latn` now parse; extlang tags such as `zh-yue` still work.

**3. Anonymous individuals are written without the `_:` prefix** (`io/ofn/writer/as_functional.rs`)

`AnonymousIndividual` was written using its raw stored label. Labels parsed from functional/Manchester input keep the `_:` prefix (the grammar's `SPARQL_BlankNodeLabel` includes it), but generated labels — e.g. the RDF reader's `anon_renumbered` (`anon000007`) — are bare, so the writer emitted `anon000007`, which is not a valid `NodeID` and fails to re-parse. Add the `_:` prefix when absent, leaving already-prefixed labels untouched.

Full lib suite green (950), including three new targeted tests.

https://claude.ai/code/session_017JttPNg3CYQnjuZ1r3Vur1

### DIFF
--- a/src/grammars/bcp47.pest
+++ b/src/grammars/bcp47.pest
@@ -12,7 +12,11 @@ BCP47_Language = ${
   | ASCII_ALPHA{5}
 }
 
-BCP47_ExtLang       = ${ ASCII_ALPHA{3} ~ ("-" ~ ASCII_ALPHA{3}){,2} }
+// Each extlang subtag is exactly 3 alpha. Without the `!ASCII_ALPHA` guards a
+// greedy PEG match would consume the first 3 letters of a following 4-alpha
+// script subtag (e.g. `zh-Hans` -> `zh` + extlang `Han`, stranding `s`); the
+// guards force extlang to yield so the script subtag can match.
+BCP47_ExtLang       = ${ ASCII_ALPHA{3} ~ !ASCII_ALPHA ~ ("-" ~ ASCII_ALPHA{3} ~ !ASCII_ALPHA){,2} }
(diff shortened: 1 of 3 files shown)


phillord:
@micheldumontier I have added you to the project. Please feel free to merge these when you are ready

PR #58, opened 2023-02-11
Create an ontology interface

phillord:
Currently, horned does not have an interface for an ontology. WIth the addition of GATs to stable rust, we should be able to add this now. 

This branch/PR is early attempts to get this working and for discussion. Not ready for merging!

PR #194, opened 2026-07-07
Add GAT-based Ontology iteration (#58)

phillord:
## Summary

- `Ontology<A>` now exposes borrowing iteration via a GAT (`type ComponentIter<'c>` + `fn iter(&self)`) and requires `IntoIterator<Item = AnnotatedComponent<A>>` as a supertrait for owning iteration, instead of the previous no-op marker trait. Implemented for all 8 current implementors (`SetOntology`, `ComponentMappedOntology`, `IRIMappedOntology`, `ConcreteRDFOntology`, and the `One`/`Two`/`Three`/`FourIndexedOntology` family, bounded so only index slot 1 needs to be iterable).
- `ForIndex` is split from one blanket impl into three concrete impls (`AnnotatedComponent<A>`, `Rc<...>`, `Arc<...>`) adding `into_component(self)`, which avoids a needless clone during owning iteration — the identity case is a pure move, and the `Rc`/`Arc` cases attempt `try_unwrap` before falling back to a clone. `unwrap` is renamed `to_component` (it was never actually unwrapping anything).
- Three benchmarks in `benches/model.rs` that deliberately isolated a single non-iterable index's raw insert cost needed reworking to call `OntologyIndex::index_insert` directly, since `MutableOntology<A>: Ontology<A>` now transitively requires iterability.
- New `benches/iteration.rs` (wired into the `horned` criterion binary) measuring borrowing vs owning iteration across synthetic classes, a real ontology fixture (`dev/family.owl`, `dev/family-other.owl` also added), and `SetOntology<RcStr>` vs `SetOntology<String>`.

Benchmarked against a `devel`-baseline worktree with the same benchmarks adapted to the pre-existing API: owning iteration (`into_iter`) is ~1.2–2.4x faster depending on type/IRI backing, borrowing iteration (`iter`) is unaffected, and every unrelated benchmark (insertion, parsing) is within normal run-to-run noise.

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo test --lib --bins --tests -- --skip integration` (1306 passed)
- [x] `cargo test --doc`
- [x] `cargo clippy --workspace`
- [x] `cargo fmt --check --`
- [x] `cargo bench --no-run` (pre-commit's Compile Benches check)
- [x] Full `bench-compare-branches` sweep against a `devel` baseline confirms the intended speedup with no regressions elsewhere


phillord:
@althonos @jannahastings @micheldumontier @b-gehrke @filippodebortoli 

I wonder whether anyone has any comments on this change. It will be (yet another) semantic change requiring a major version; done a few of those lately. But I've felt for a long time having not trait interface for `Ontology` at all was poor. This makes explicit that ontology is an iterable collection of `AnnotatedComponent`s. 

Not done lots of GAT based stuff (i.e. "not lots" in the "any at all" sense). I am not seeing any downsides.

micheldumontier:
makes sense. go for it.

Issue #196, opened 2026-07-08
Two #[ignore]d tests are broken: fix fixtures/paths or remove them


phillord:
Auditing the 4 currently `#[ignore]`d tests by running each with `--ignored`, two do not actually pass:

- `horned-bin/tests/test_horned_materialize.rs:20` (`integration_ont_with_bfo`) — expects a fixture at `devel/tmp/ont-with-bfo.owl` which does not exist, so it fails with `IOError(NotFound)` before it ever exercises the BFO-import/network behaviour it's meant to test.

- `tests/manchester/corpus.rs:233` (`corpus_parses_or_documents_blocker`) — `corpus_paths()` hardcodes fixture paths under `/data/dumontier/pymos/bench/data/{koala,sio,obi-core,hp}.rdfxml`. That directory doesn't exist outside whatever machine those paths were captured on, so the path list is empty and `assert!(!rows.is_empty(), "no corpus fixtures found")` panics. Note this is a genuine assertion failure, not the "skips gracefully when docker is absent" behaviour documented in the module doc — docker was present and reachable when this was tested, the corpus-fixture lookup fails before docker is ever invoked.

For comparison, the other two ignored tests are fine as-is:
- `tests/manchester/report.rs:11` (`generate_compliance_report`) passes cleanly.
- `tests/manchester/canonical.rs:328` (`corpus_axiom_equality_documents_diffs`) hits the same missing `/data/...` fixtures as corpus.rs but has an early-return guard (`if rows.is_empty() { ...; return; }`) instead of an assert, so it exits cleanly rather than failing.

Either fix these two (regenerate/relocate the missing fixtures, or make the corpus path list configurable/discoverable rather than hardcoded to one machine's layout) or remove them if they're no longer worth maintaining.

micheldumontier:
that is very silly. will fix.

micheldumontier:
Fixed both broken tests in #216.

- `integration_ont_with_bfo`: made hermetic (stages the in-repo `src/ont/owl-rdf/ont-with-bfo.owl` into a temp dir, runs, cleans up) instead of relying on the missing `../tmp/` fixture. Also corrected a latent wrong assertion — the BFO import localizes to `obo_bfo.owl`, not `bfo.owl`; the old check never ran because the test died on the missing fixture first. Still `#[ignore]`d for network; verified passing against a real BFO fetch.
[...]

PR #216, opened 2026-07-14
Fix two broken #[ignore]d tests (fixtures/paths)

micheldumontier:
Fixes the two broken `#[ignore]`d tests identified in #196.

## `integration_ont_with_bfo` (horned-bin)
Previously expected a fixture at `../tmp/ont-with-bfo.owl` that no longer exists, so it failed with `IOError(NotFound)` before ever exercising import materialization.

- Made hermetic: stages the in-repo fixture (`src/ont/owl-rdf/ont-with-bfo.owl`) into a fresh per-process temp dir, runs there, and cleans up. No reliance on a hand-placed `tmp/` fixture and no work-tree pollution (the old version wrote `bfo.owl` into the shared `tmp/` and never cleaned up, so it could only pass once).
- Fixed a latent wrong assertion: the BFO import (`http://purl.obolibrary.org/obo/bfo.owl`) localizes to **`obo_bfo.owl`** (underscore-joined IRI path), not `bfo.owl`. The old check never ran because the test died on the missing fixture first.
- Still `#[ignore]`d (requires network). Verified passing with a real BFO download.

## `corpus_parses_or_documents_blocker` (manchester conformance)
`corpus_paths()` hardcoded fixtures under `/data/dumontier/pymos/bench/data`, so off that machine the path list was empty and `assert!(!rows.is_empty())` panicked — a genuine assertion failure, not the graceful docker-absent skip the module doc claims.

- Read the corpus dir from `HORNED_CORPUS_DIR` instead of a hardcoded path; return empty when unset/absent.
- Skip the test gracefully (with a message) when no corpus is configured.
- `canonical.rs` needed no change — it already routes through `corpus::corpus_paths()` with an early-return guard, so the central fix covers it.

## Verification
- `integration_ont_with_bfo` — passes with `--ignored` (real network fetch).
- `corpus_parses_or_documents_blocker` — skips cleanly and passes (docker present, corpus absent).
- Default suites still green: `test_horned_materialize` (1 passed, 1 ignored), `manchester_conformance` (20 passed, 3 ignored).

Note: the corpus fixtures (`koala`, `sio`, `obi-core`, `hp`) are genuinely gone and not recoverable from the original machine, so making the path configurable is the right call rather than relocating them.

### DIFF
--- a/horned-bin/tests/test_horned_materialize.rs
+++ b/horned-bin/tests/test_horned_materialize.rs
@@ -1,7 +1,7 @@
 use assert_cmd::cargo;
 use assert_cmd::prelude::*; // Add methods on commands
 use predicates::prelude::*; // Used for writing assertions
-use std::{path::Path, process::Command}; // Run programs
+use std::{fs, path::Path, process::Command}; // Run programs
 
 #[test]
 fn integration_run() -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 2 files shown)


phillord:
Could you squash these two, otherwise happy to rebase

micheldumontier:
@phillord Squashed into one commit.

Issue #237, opened 2026-07-19
RDF/XML round-trip fails for entities used in class position but never locally declared (cross-import punning)


phillord:
## Summary

`horned-roundtrip` corpus case `HO` (`ofn -> rdf_xml`) fails with a
`reread_fail`. Filing this for the record, but I believe it is **not** a
narrow horned-owl bug in the same sense as #226-#236 -- see "Why I'm not
fixing this" below.

## Root cause

`HO`'s source OFN declares `Import(<https://w3id.org/def/Organ>)` /
`Import(<https://w3id.org/def/Tissue>)` and then uses classes from those
imports (e.g. `Organ:OrganoVena`) directly in axioms, without a local
`Declaration(Class(Organ:OrganoVena))` -- the declaration lives in the
(unfetched, since this is a closed round-trip test with no import
resolution) imported ontology. horned-owl's OFN reader is lenient about
this: `SubClassOf(Organ:OrganoVena ...)`'s grammar position
unambiguously tells the parser `OrganoVena` is used as a class, no
declaration lookup needed. I confirmed via a small script dumping the
parsed model that no `DeclareClass` component for `OrganoVena` exists
anywhere in it -- only the three axioms that use it.

When that model is written to RDF/XML and reread, `owl:equivalentClass`
is ambiguous in the OWL2 RDF mapping (it's used both for
`EquivalentClasses(class, class)` and `DatatypeDefinition(datatype,
datarange)`), so `io::rdf::reader::distinguish_term_kind` must look up
the subject's *declared* kind to disambiguate -- and finds none, since
(as above) the model never had one. That's exactly the same class/kind
lookup that `distinguish_retrieve_property_kind` already special-cases
with a `_ if self.config.lax => ...` fallback a few lines away in the
same file; `distinguish_term_kind` has no equivalent lax fallback, so it
falls through to the generic "Unknown entity in equivalent class
statement" error -- and `horned-roundtrip`'s harness reads with the
default (non-lax) `ParserConfiguration` on both sides anyway, so even
adding a lax-gated fallback wouldn't fix this specific corpus case.

Minimal repro: `Class: A`, `SubClassOf(A ...)`,
`EquivalentClasses(A ObjectSomeValuesFrom(...))` with **no**
`Declaration(Class(A))` anywhere in the source OFN -> write to RDF/XML ->
reread fails the same way.

## Why I'm not fixing this

Unlike #226-#236, the model here is faithfully round-tripping what it was
[...]

Issue #236, opened 2026-07-19
bcp47.pest: BCP47_Script/BCP47_Region greedily truncate a longer variant subtag (e.g. en-scotland fails to parse)


phillord:
## Summary

`horned-roundtrip` corpus case `FOODON` (`rdf_xml -> omn`) fails with a
`reread_fail`: the OMN writer emits a BCP-47 language tag verbatim (e.g.
`@en-scotland`), and horned-owl's own `bcp47.pest`-based Manchester Syntax
reader then fails to parse that same, valid, tag back.

## Root cause

`src/grammars/bcp47.pest`'s `BCP47_Script` and `BCP47_Region` productions
match a *fixed*-length prefix of the input with no trailing
subtag-boundary guard:

```
BCP47_Script        = ${ ASCII_ALPHA{4} }
BCP47_Region        = ${ ASCII_ALPHA{2} | ASCII_DIGIT{3} }
```

For a tag like `en-scotland` (language `en` + the single 8-character
variant subtag `scotland`, no script/region present), the PEG engine
greedily matches the *first four letters* of `scotland` (`scot`) as an
optional `BCP47_Script`, since nothing requires the subtag to be followed
by `-` or end-of-tag. That leaves `land` stranded with no leading `-`, so
none of the following optional groups (region/variant/extension/private
use) can consume it, and the overall `BCP47_LangTag` rule succeeds having
consumed only `en-scot`, with `land` left over -- which then fails to
parse at the outer (`Annotations`/`Frame`) level.

This is the exact same class of PEG greedy-match bug already fixed for
`BCP47_ExtLang` (see the `!ASCII_ALPHA` guards and comment already present
on that rule) -- the fix was just never applied to `BCP47_Script` /
`BCP47_Region`, which are equally vulnerable.

## Minimal reproduction

```
Prefix: : <http://ex/>
Prefix: rdfs: <http://www.w3.org/2000/01/rdf-schema#>
Ontology: <http://ex/o>
Class: :A
    Annotations: rdfs:label "neep"@en-scotland
```

`omn::reader::read` fails with:

```
expected EOI, GeneralAxiomBlock, Annotations, Frame, or Misc
   |
   Annotations: rdfs:label "neep"@en-scotland
   |                                   ^--- (right after "en-sco")
```

(`en-scotland` is a real tag found in FOODON's RDF/XML source, on an
`oboInOwl:hasExactSynonym "neep"@en-scotland` annotation; the OMN writer
[...]

Issue #235, opened 2026-07-19
OFN writer emits ObjectIntersectionOf/ObjectUnionOf with <2 operands, producing unparseable output


phillord:
## Summary

`horned-roundtrip` corpus case `BCS7` (`turtle -> ofn`) fails with a
`reread_fail`: horned-owl writes syntactically invalid OWL Functional
Syntax when the model contains an `ObjectIntersectionOf`/`ObjectUnionOf`
with only a single operand, and its own OFN reader then rejects that
output.

## Root cause

`ofn.pest`'s grammar requires at least two operands:

```
ObjectIntersectionOf   = { LIT_OBJECT_INTERSECTION_OF ~ LBRACKET ~ ClassExpression{2, } ~ RBRACKET }
ObjectUnionOf          = { LIT_OBJECT_UNION_OF ~ LBRACKET ~ ClassExpression{2, } ~ RBRACKET }
```

`ClassExpression::ObjectIntersectionOf`/`ObjectUnionOf` in `model.rs` are
just `Vec<ClassExpression<A>>` with no arity check, and the RDF reader
happily builds a single-operand `ObjectIntersectionOf` from a malformed
(but real-world) `owl:intersectionOf` RDF list that only has one member --
e.g. `BCS7`'s source Turtle contains:

```turtle
[] a owl:Class ;
    rdfs:subClassOf cst:R7_Stage_IV ;
    owl:intersectionOf ( cst:M1 ) .
```

RDF/XML and OWL/XML round-trip this fine (their grammars don't enforce a
minimum operand count), but the OFN writer
(`src/io/ofn/writer/as_functional.rs`, the `ObjectIntersectionOf`/
`ObjectUnionOf` arms) writes it out verbatim as
`ObjectIntersectionOf(<...M1>)`, which its own `ofn.pest`-based reader then
rejects:

```
Parsing Error: expected ClassExpression
   |
   SubClassOf(ObjectIntersectionOf(<...#M1>) <...#R7_Stage_IV>)
   |                                                          ^--- (needs a 2nd ClassExpression)
```

## Minimal reproduction

Read the corpus's `BCS7` Turtle file, write it to OFN, and reread the OFN:
the reread fails. (Full corpus file: `BCS7.gz` in the `horned-roundtrip`
corpus; the relevant triple is the `owl:intersectionOf ( cst:M1 )` shown
above, which parses into a single-operand `ObjectIntersectionOf` and is
used as the superclass of a `SubClassOf` axiom.)

[...]

Issue #234, opened 2026-07-19
OFN writer emits unescaped invalid IRI characters (e.g. '[', ']'), breaking reread


phillord:
The OFN writer's IRI-emitting code (`Display for Functional<'_, IRI<A>, A>` and `Display for Functional<'_, curie::PrefixMapping, A>` in `src/io/ofn/writer/as_functional.rs`) writes the underlying IRI text straight into the `<...>` full-IRI form with no validation or escaping.

horned-owl's `IRI<A>` is just an interned string; readers for XML- and RDF-based formats (in particular `src/io/owx/reader.rs`) hand IRI attribute/node text straight through without validating it's a legal RFC 3987 IRI, since that's not required by those formats' own grammars. Real-world OWL/XML ontologies exploit this: e.g. `MCCL` (from the horned-roundtrip corpus) declares:

```xml
<Prefix name="R" IRI="http://www.semanticweb.org/pallabi.d/ontologies/2014/2/untitled-ontology-11#KB-CH[R]-8-5"/>
```

`[` and `]` are legal, unescaped in an XML attribute value, so the OWX reader accepts this fine and the resulting IRI string carries the literal `[`/`]` through into the model. But per `src/grammars/rfc3987.pest`, `[` and `]` are `RFC3987_IriGenDelims`, legal only inside the authority's `IP-literal` production -- not in a fragment (see `RFC3987_IriFragment`, which only permits `RFC3987_IriIpChar | "/" | "?"`). So when the OFN writer emits this IRI as `<...KB-CH[R]-8-5>`, the OFN grammar's own `FullIRI` rule rejects it on reread.

## Minimal reproduction

Using `MCCL`'s OWL/XML source (`horned-roundtrip` corpus), read via `owx::reader::read`, write via `ofn::writer::write` (passing the recovered `PrefixMapping`), then reread the output with `ofn::reader::read`:

```
read ok, 19083 components
wrote ... bytes
REREAD FAILED: Parsing Error:  --> 2:93
  |
2 | Prefix(R:=<http://www.semanticweb.org/pallabi.d/ontologies/2014/2/untitled-ontology-11#KB-CH[R]-8-5>)
  |                                                                                             ^---
  |
  = expected RFC3987_IriPctEncoded or RFC3987_IriUCSChar
```

## Root cause

[...]

PR #238, opened 2026-07-20
Fix 12 horned-roundtrip corpus reread_fail regressions

phillord:
## Summary

Twelve independently-verified round-trip bugs found via `horned-roundtrip` corpus testing, each fixed on its own branch and merged here:

- #226 — OWL/XML reader: doubled `#` in `<IRI>text</IRI>` element content
- #227 — OWL/XML writer: invalid `xmlns:="..."` for the empty/default prefix
- #228 — order-dependent `AnnotationProperty`/`ObjectProperty` pun resolution
- #229 — OFN writer: `Import` written after ontology `Annotation`
- #230 — OFN writer: invalid `AbbreviatedIRI` for the default/empty CURIE prefix
- #231 — OFN writer: abbreviates IRIs with an invalid `PN_LOCAL` local name
- #214 — OFN writer: degenerate single-member `SameIndividual`/`DifferentIndividuals`
- #232 — RDF/XML writer: unescaped RFC 3987-invalid IRI characters
- #233 — OMN writer: invalid bare `SimpleIRI` for curie's default-slot prefix match
- #234 — OFN writer: unescaped RFC 3987-invalid IRI characters
- #235 — OFN writer: degenerate single-operand `ObjectIntersectionOf`/`ObjectUnionOf`
- #236 — BCP-47 grammar: greedy script/region match truncating long variant subtags

Each fix has its own regression test, confirmed to fail without the fix and pass with it.

## Impact

Full corpus round-trip (`reread_fail` outcome): **447 → 36** (92% reduction), zero new panics/read_fail/write_fail throughout. One case (`HO`, issue #237) was investigated and correctly left unfixed — it's a genuine missing-declaration gap in the source ontology, not a narrow encoding bug, and forcing a fix would require a disproportionately large change.

## Performance

Benchmarked against `devel` (`bench-compare-branches`). Reading, iteration, and in-memory structures are unaffected (within ~0-5% noise). RDF-family writers (RDF/XML, Turtle, N-Triples) show a real, consistent slowdown (+8-27% depending on format) from the new IRI-escaping check added on the shared IRI-to-term conversion path (#232) — small in absolute terms (sub-millisecond even at 10,000 axioms) and understood as the direct cost of producing valid, rereadable output for previously-broken cases.

## Test plan

- [x] Each fix's own regression test verified red→green
- [x] Full workspace `cargo test --lib --bins --tests -- --skip integration` clean (1371 tests)
- [x] `cargo clippy --workspace --all-targets` clean
- [x] `cargo fmt --check` clean
- [x] Full corpus round-trip rerun confirming the 447→36 reduction with zero regressions

### DIFF
--- a/src/grammars/bcp47.pest
+++ b/src/grammars/bcp47.pest
@@ -17,8 +17,18 @@ BCP47_Language = ${
 // script subtag (e.g. `zh-Hans` -> `zh` + extlang `Han`, stranding `s`); the
 // guards force extlang to yield so the script subtag can match.
 BCP47_ExtLang       = ${ ASCII_ALPHA{3} ~ !ASCII_ALPHA ~ ("-" ~ ASCII_ALPHA{3} ~ !ASCII_ALPHA){,2} }
-BCP47_Script        = ${ ASCII_ALPHA{4} }
-BCP47_Region        = ${ ASCII_ALPHA{2} | ASCII_DIGIT{3} }
+// Script and region are fixed-length subtags (4 alpha / 2 alpha or 3
+// digit), but that length is only a *minimum* match for a PEG engine --
+// without a trailing `!ASCII_ALPHANUMERIC` guard, a longer, unhyphenated
+// variant subtag right after the language (e.g. the 8-alphanumeric variant
+// in `en-scotland`) gets wrongly truncated: `scot` matches `BCP47_Script`
+// and the remaining `land` is left stranded with no `-` before it, so the
+// whole tag fails to parse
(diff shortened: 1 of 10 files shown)


phillord:
Reverted: this merged the whole integration branch (12 fixes) into `devel` when only `fix/226-owx-iri-element-double-hash` had actually been reviewed at the time. Force-pushed `devel` back to `30ab61c` (its state immediately before this merge). The other 11 fixes remain available on their own branches for individual review; this PR stays closed/merged as a historical record but its commits are no longer on `devel`.

PR #242, opened 2026-07-20
Comparison with EBI fork

jamesamcl:
## Summary

This branch is the EBI ([EBISPOT](https://github.com/EBISPOT/horned-owl)) fork's `devel`, opened as a comparison against upstream `phillord:devel`. It carries six commits of correctness, performance, interoperability, and WebAssembly work, driven by using horned-owl to read and write the OWL that ROBOT / the OWLAPI produce over large ontologies.

The changes fall into four themes. Serialization output stays stable throughout: each set is sorted before writing, so the hash-function switch below doesn't perturb ordering, and all `io::ofn` tests pass.

### 1. RDF/OFN reader & writer fidelity

- **Non-destructive class-expression retrieval** — a class-expression blank node referenced by more than one parent is now cloned on retrieval instead of removed on first use, so shared nodes are no longer lost.
- **Keep every reified annotation set** per base triple (previously only the last survived); type literal assertions on undeclared properties as annotation assertions, matching ROBOT.
- **Canonical operand ordering** for intersections/unions, so a structurally equal collection matches regardless of which blank node carries the list.
- Reified-annotation and n-ary `disjoint` round-tripping; SWRL rule annotations written on the rule node; OFN writer UTF-8 slicing and class-expression rendering fixes.

### 2. Performance on the hot paths

- **FxHash (rustc-hash) instead of SipHash** on the IRI / anonymous-individual interning caches, the component/triple stores, the RDF reader's blank-node maps, and the declaration / logically-equal indexes. Output is unchanged because each set is sorted before serialization.
- **O(N) RDF-list stitching** — each `rdf:List` is assembled in a single pass over its cells, instead of re-scanning every blank node once per element (previously O(list-length × bnode-count), pathological on large ontologies).
- Pre-size the `SetOntology` `HashSet` and build it by *moving* components out of the index rather than cloning.

### 3. ROBOT / OWLAPI functional-syntax compatibility

- **Grouped OFN writer output** — `io::ofn::writer::write` now emits the layout the OWLAPI (and hence ROBOT / dosdp-tools) produces: a canonical prefix block, an ontology header, a leading block of `Declaration(...)` grouped by entity type and sorted by IRI, then `# Class: <IRI> (label)` sections carrying each entity's annotation assertions followed by its logical axioms. This makes output byte-comparable with committed ROBOT `.ofn` files. The `#` comments are skipped by the OFN reader, so output still round-trips.
- **LANGTAG parse fix** — use the concrete-syntax `LANGTAG` production shared by Turtle / SPARQL / OWL 2 FS instead of the full BCP47 grammar, which as a PEG greedily mis-parsed a well-formed tag like `zh-hans` as `zh-han` + a stray `s`.
- **Bare anonymous node ids** — also accept the `anon000001`-style ids OWLAPI/ROBOT emit for anonymous individuals, alongside the spec's `_:`-prefixed blank-node label, guarded so the bare form can't swallow an abbreviated IRI or a functional keyword.

### 4. WebAssembly support (wasm32 + wasm64)

- A new `crate::time` shim replaces the direct `Instant::now()` calls in the RDF reader and `SetOntology` build (used only for optional perf timing behind a timing env var). On `wasm32` the std clock traps, so merely reading it aborts the module — which otherwise makes browser RDF/XML parsing impossible. The shim routes through `web-time` on wasm32, a host-imported monotonic clock (`host.now_nanos`) on the wasm64 reactor target, and `std::time` natively. No behavioural change off wasm.
- `indexmap` gains `features = ["std"]` so it builds on tier-3 `wasm64-unknown-unknown` (where its autocfg sysroot probe would otherwise compile it no-std and break the build). Harmless elsewhere.

## Known issue

`horned-bin`'s `integration_validate_ontology_rdf` now fails: `horned-validate` treats any entry left in the parser's incomplete pool as a failed parse, and the non-destructive class-expression retrieval in theme 1 intentionally leaves shared blank nodes there. The parse itself is correct — only the leftover-as-incomplete heuristic is tripped.

## Scope

15 files changed, +950 / −207. New dependencies: `rustc-hash` (all targets) and `web-time` (wasm32 only).

## Commits

1. RDF/OFN reader+writer fidelity fixes, O(N) list stitching, and FxHash on the hot paths
2. io/rdf, ontology/set: use a wasm-safe clock (web-time)
3. time: build and run on wasm64 (host clock; indexmap std)
4. Emit ROBOT/OWLAPI-style grouped functional syntax from the OFN writer
5. OFN reader: accept W3C LANGTAG and ROBOT/OWLAPI bare anonymous node ids


phillord:
I notice that you are also using a vendor branch for horned in hermit-rs (which is a bit behind release now). Do these overlap?

jamesamcl:
Good catch

hermit-rs's vendored horned is an older subset of this PR - have checked with claude. So hermit-rs can switch back to upstream horned once this is reconciled.

phillord:
I have a bunch of ofn/rdf commits going in at the moment, so we may be working on the same thing.

phillord:
Would suggest that you take this out of draft and we can go for a merge/rebase.

I'm also working on the code base and the conflicts are just going to get worse over time!

PR #284, opened 2026-08-13
Add horned-corpus: round-trip, profile and reasoning conformance harness

phillord:
## Summary

- Subtree-imports the standalone `horned-roundtrip` repo (Michel Dumontier) as a new workspace member, `horned-corpus`, dual-licensed MIT/LGPL-3.0. Full original history is preserved on the merge commit's second parent.
- Splits its old combined `run` command into three independent sweeps over a corpus, each streaming JSONL: `roundtrip` (read/write/reread + diff categorization), `profile` (OWL 2 EL/QL/RL/DL conformance via `horned-profile`, optionally cross-validated against ROBOT's `validate-profile`), and `reason` (ELK/HermiT/JFact via ROBOT — the Rust reasoners can't be wired in yet, they all pin `horned-owl ^1.4`).
- `report` aggregates any of the three into `cases.csv` / `summary.json` / `report.md`, rendering only the sections the record type supports.
- ROBOT is now pinned and auto-downloaded (`robot::robot_ensure`, mirroring horned-owl's own `bubo_ensure`) rather than resolved off `$PATH`, so `reason` and `--robot-ground-truth` runs are reproducible against a known release (v1.9.10) instead of whatever happens to be installed.
- Adds Michel Dumontier to `MAINTAINERS.md` for `horned-corpus`.

Validated against the real ~1189-file BioPortal corpus: full `roundtrip` and `profile` sweeps completed clean, and a full `reason` sweep (ELK + HermiT, 15 min/ontology cap) finished with no crashes.

## Test plan

- [x] `cargo test` passes (workspace + horned-corpus's own 51 unit tests)
- [x] `cargo fmt --check` / `cargo clippy` clean
- [x] Full corpus `roundtrip`, `profile` and `reason` sweeps run end-to-end without panics
- [x] `robot_ensure()` verified to download and run ROBOT v1.9.10 correctly from a clean `dev/`

**Not merging yet** — opening for review/visibility only.


phillord:
@micheldumontier this is my attempt to integrate this directly into horned owl. Are you happy with this?

micheldumontier:
@phillord Yes, but a couple of things:
- robot.rs downloads the ROBOT jar (robot.jar) by via wget which isn't on stock macOS/Windows. Since we already depend on reqwest (rustls-tls) for the BioPortal fetcher, we could pull the jar through that instead and drop the subprocess — pure-Rust, no external binary. Same applies to bubo_ensure.
- cargo clippy flags two collapsible_if in fetch.rs on 1.96+ toolchains inherited from my original code

phillord:
Good point. We've fixed robot_ensure. I'll do bubo_ensure at some other point in time, since horned_owl uses ureq rather than reqwest.

PR #291, opened 2026-08-27
Rework of read/write interfaces, adding a streaming option

phillord:


phillord:
Comments welcome on this one as it is a fairly significant change, but hopefully regularizes all of the interfaces for parsers.

phillord:
@balhoff @jannahastings @KonradHoeffner @micheldumontier @althonos

jannahastings:
@b-gehrke

KonradHoeffner:
Thanks for the ping, I can't comment much, but will document migration of my own code. I am already using horned-owl as git dependency and changed the branch from devel to this one, getting the following error on first compile:

```
error[E0425]: cannot find function `read_with_build` in module `horned_owl::io::rdf::reader`
    --> src/rdf.rs:229:106
     |
 229 | ...o::rdf::reader::read_with_build(
     |                    ^^^^^^^^^^^^^^^ not found in `horned_owl::io::rdf::reader`
note: similarly named function `parser_with_build` defined here
```

Alright, let's just change "read_with_build" to "read", drop the "build" and change "ParserConfiguration" to "RDFParserConfiguration" and it builds and runs again, very painless from my point of view, so thumbs up! Sorry if I can't give you deeper feedback.

```
-    Ok(horned_owl::io::ParserOutput::<Arc<str>, ArcAnnotatedComponent>::rdf(horned_owl::io::rdf::reader::read_with_build(
+    Ok(horned_owl::io::ParserOutput::<Arc<str>, ArcAnnotatedComponent>::rdf(horned_owl::io::rdf::reader::read(
         &mut br,
-        &b,
-        ParserConfiguration::default(),
+        RDFParserConfiguration::default(),
     )?)
     .decompose()
     .0)
```

b-gehrke:
Now that the format is in the configuration I wonder if it makes sense to expose a `read` function at the `io` module level which matches on the input format and selects the proper parsing function. Similarly, a `write` function could be added but I think we don't have a struct for writer configurations.

I feel like Functional Syntax should also be possible to be parsed as a stream, wouldn't it? Unfortunately, pest doesn't support that and we would have to switch to another parser library or come up with some workarounds (like counting parenthesis). 

On a side note, I'm a bit confused about the `InputType` and `ResourceType` enums since the former seems to entails the latter.

phillord:
@b-gehrke Apologies for missing you off the @ links. Github doesn't have an "at all" option as far as I can see, which is a bit irritating.

Yes, I have added all the support for format guessing into horned_owl::io, mostly stolen from horned-bin, stolen from @micheldumontier. We should uncover a "read(BufRead)" option to io directly.

As you note pest breaks stream parsing for OMN and OFN which is a bit unfortunate. Simple roundtrip tests of OWX show a 20-30% speed up and, of course, much lower memory usage, so it's a gain worth having. The framework is there now, and rust does have some streaming parsers, so something for the future.

I will work out what is happening with InputType and ResourceType.

b-gehrke:
Yes, the framework and the performance and memory improvements look great indeed!

I'm also looking into other parses at the moment to add support for ROBOT like "labels as IRIs".

phillord:
@b-gehrke nom stream parses. We could add one for omn or ofn along side the existing one while we do a compare and contrast. The test framework, combined with horned-corpus is pretty good now, which would give a lot of safety,

balhoff:
@phillord I have been working on a new streaming-capable OFN parser, so this is great timing! It's about 5x faster than the current OFN parser for Uberon and GO-Plus. I would be happy to have this absorbed into horned-owl if you are interested.

https://github.com/balhoff/fast-ofn

phillord:
@balhoff That is a lot faster than I would have expected. Yes, very happy to incorporate this into horned. We would need to think about whether to add it to or replace the existing parser. If you are happy I will set up a scratch branch on horned, so I can see how well it works alongside the other code.

I have some thinking to do about horned-owl organisation. It was single crate originally and hence all the IO is there but this does mean it keeps getting bigger.

